PR #1, opened 2013-06-28
M2Crypto libraries fix

vladum:
LD_RUN_PATH is only used at link time to set RPATH in ELF - this is only needed for M2Crypto, but I used `setup.py build_ext --rpath=...` which does the same thing. LD_PRELOAD for libcrypto shouldn't be set because it affects ssh which in turn is used by prun.


vladum:
Please wait with this one. Seems like the full build is broken, though rebuilding M2Crypto manually worked.

vladum:
Seems fine now.

PR #14, opened 2013-07-16
initial version of spectraperf

corpaul:


spectraperf/DatabaseHelper.py
```diff
@@ -0,0 +1,93 @@
+'''
```
whirm: Lets use lowercase for all the files

scripts/run_in_env.sh
```diff
@@ -44,6 +44,9 @@ echo "Project root is: $PROJECTROOT"
 # Load env variables for this experiment
 source experiment_vars.sh
 
+# Update PYTHONPATH
+export PYTHONPATH=$PYTHONPATH:$PROJECTROOT/spectraperf
```
whirm: Better to add $PROJECTROOT to the python path and update the module imports. That way we avoid name conflicts between projects.

tests/testPerformanceProfile.py
```diff
@@ -0,0 +1,179 @@
...
+        #p.addToRange("test2", 10)
+        #p.addToRange("test2", 20)
+
+
+        fits = p.fitsProfile(sess1)
+        self.assertEqual(fits["test1"], 1)
+
+        sess1.addStacktrace(s2)
+        # should raise AssertionError because we did not add a range for test2 yet
+       # with self.assertRaises(AssertionError):
```
whirm: bad indentation levels

scripts/run_in_env.sh
corpaul: where are those and what should I change?

PR #39, opened 2013-08-15
Similarity report

corpaul:
stuff used to calculate the activity matrices for all revisions, including the number of offBytes and rangeDiff


whirm:
Should we merge this?

whirm:
yup.

PR #44, opened 2013-10-10
Similarity report

corpaul:
all code needed to generate the reports as used in our case study for the paper
(so suitable for batch processing at the moment, the scripts should be updated to work for 'realtime' reporting)


corpaul:
the stuff should be OK for per-revision processing now as well

whirm:
Closing this one as it's empty

PR #67, opened 2013-11-22
Documentation improvements

whirm:
Also, implement a script (generate_config_file.py) that parses all of Gumby's resources and generates a config template with all the available documentation, scripts and configuration parameters available.

There's still some stuff pending to document, but we are getting there.


NielsZeilemaker:
Looking mighty impressive

PR #74, opened 2013-11-29
Removed manual head_node configuration

NielsZeilemaker:


experiments/dispersy/allchannel.conf
```diff
@@ -31,9 +31,6 @@ das4_reserve_duration = 1000
 # How many processes do we want to spawn?
 das4_instances_to_run = 1000
 
-# We need to manually set this on the das4 as dispersy will get really confused otherwise
```
whirm, line 34: I would leave some comment telling people about it so they don't go crazy when trying to run the experiment from outside of Jenkins.

NielsZeilemaker:
The new dispersy is supposed to not need this anymore.  So, I simply removed them.

whirm:
Oh, so you aren't exporting it from jenkins? That's nice!

NielsZeilemaker:
I am,  but the new dispersy probably doesn't require it. But that version
hasn't been merged into development yet.

PR #86, opened 2013-12-04
fix for stacktrace on empty files

corpaul:


whirm:
Shouldn't we fail if we get asked to graph an empty file?

corpaul:
No it traverses all the files in the output directory right... for me it's quite useful if it doesn't fail, so that for example the seeder doesnt write resource usage at least I don't have to repeat the whole experiment

whirm:
Yeah, I guess you are right.

PR #104, opened 2014-01-02
Synchronize the start of the experiment by scheduling it to a global

### LINKED ISSUE #101: Problems with GO command
Currently the SyncServer sends a GO command to all clients ordering them to start the experiment. However, this results in two problems:
- The GO command will not arrive at all clients at the same time -> hence peers start at different times.
- Only after receiving the GO command is the scenario file parsed, causing huge differences between nodes. In one run I could see more than a 2s difference between actually starting the experiment.

Suggested changes:
- Add a timestamp to the GO command indicating when a client should start
- Before sending the vars, the scenario file should be parsed, after receiving the GO command + timestamp the tasks should be registered.

whirm:
time and taking into account the local clock differences between instances.

fixes #101

### DIFF
--- a/gumby/sync.py
+++ b/gumby/sync.py
@@ -21,7 +21,7 @@
 #
 # When the all of the instances we are waiting for are all ready, all the information will
 # be sent back to them in the form of a JSON document. After this, a "go" command will
-# be sent to indicate that they should start running the experiment.
+# be sent to indicate that they should start running the experiment with the absolute time at which the experiment should start.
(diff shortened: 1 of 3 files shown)


whirm:
Oops I forgot to remove the call delay, 1 sec.

whirm:
ok, it should be fine now

NielsZeilemaker:
Great, I'll update my gummy branch to test it

Issue #106, opened 2014-01-02
Clients are not reconnecting


NielsZeilemaker:
See 
http://jenkins.tribler.org/job/Experiment_PrivateSocial_Act/COMMUNITY_TYPE=poli,DURATION=600,NODES=40,PEERS=4002,label=das4_vu/74/console
and
http://jenkins.tribler.org/job/Experiment_PrivateSocial_Act/COMMUNITY_TYPE=poli,DURATION=600,NODES=40,PEERS=4002,label=das4_vu/73/console

Gumby clients are not reconnecting to the sync server whenever it does not respond.

whirm:
I've tried the devel branch several times with 6000 instances and it does retry. Ill continue investigating. Maybe we need to lower the retry time ceiling.

NielsZeilemaker:
Maybe it's my change of parsing the scenario file before sending the ready
command

whirm:
I managed to trigger it every once in a while with 10000 clients but the server starts doing really strange things like taking minutes to die, so I'm not sure it's the same situation as yours, can you try with this pull request?

NielsZeilemaker:
My feeling is that maybe it has to do with my parsing of the scenario file,
that is taking roughly 12s per peer. I'm going to offload it to another
thread using deferToThread and see if that helps.
Niels

2014/1/3 Elric Milon notifications@github.com

whirm:
Take into account that anything you do reactor-related has to be called in the reactor thread (reactor.callFromThread())

https://twistedmatrix.com/documents/12.2.0/core/howto/threading.html

NielsZeilemaker:
I'm currently only parsing the file, not scheduling the calls so that
should be fine.

2014/1/3 Elric Milon notifications@github.com

whirm:
@NielsZeilemaker Is this still happening to you?

NielsZeilemaker:
I think it is

whirm:
I hasn't happened to me in these two months, how is it for you?

PR #107, opened 2014-01-02
Limit the time between reconnection attempts

whirm:


whirm:
I tested it a bunch of times with 10000 nodes and it works OK

Should fix #106

PR #108, opened 2014-01-06
Run in env cleanup

whirm:


whirm:
I've tested it with a few dummy experiments and it ran ok, @NielsZeilemaker anything against merging it?

NielsZeilemaker:
Seems fine, i'll merge it

PR #109, opened 2014-01-07
Client connection fixes

whirm:


whirm:
I tested it a couple dozen times with 10000 and it succeeded greatly all of them. (The experiment server still needs 5 minutes to shut down though)

PR #129, opened 2014-02-14
Refactor for get_member() changes in dispersy

whirm:
This needs to be merged at the same time than https://github.com/Tribler/dispersy/pull/233

### DIFF
--- a/gumby/experiments/dispersyclient.py
+++ b/gumby/experiments/dispersyclient.py
@@ -244,13 +244,18 @@ class DispersyExperimentScriptClient(ExperimentClient):
 
         self._dispersy.start()
 
-        self._master_member = self._dispersy.callback.call(self._dispersy.get_member, (self.master_key, self.master_private_key))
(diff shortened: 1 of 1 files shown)


whirm:
@NielsZeilemaker Ready!

NielsZeilemaker:
is this change backwards compatible?

whirm:
It won't work with an old dispersy if we pass the private key only.
Do you want to keep backwards compatiblity here? It would only matter if we where to run run das4 experiments with a new gumby and an old dispersy.

https://github.com/Tribler/dispersy/pull/233/files#diff-6ca8135667b3a3cb278e4383ff45d89cL522

Before we where forced to pass the public key even when passing the private/public keypair

NielsZeilemaker:
You're basically describing all jenkins experiments we have except yours, allchannel, demers, etc are all using gumby/devel + tribler/devel + dispersy/next

whirm:
I've undone the merge, let's talk about it when I'm back

PR #132, opened 2014-02-20
Unit stap

corpaul:
stuff to run the tribler unittests in stap (including the stuff to fix the virtual env)


whirm:
Could you squash/drop the superfluous commits?

corpaul:
this better?

PR #151, opened 2014-03-27
Fix data index and metadata experiment

LipuFei:


LipuFei:
@NielsZeilemaker @whirm Ready

NielsZeilemaker:
Whirm if you agree with the - 6 stuff while parsing the log files,  please merge it

whirm:
Yeah, we need to do it like this for IO stats as they are always appended at the end and stat file can have a different amount of values depending on the kernel.

PR #171, opened 2014-07-16
Implement manhole support for gumby experiments

whirm:
Also add meliae support


whirm:
@corpaul  ready to merge

Also, If you could add the config params to the Tribler 1h, 24h, 1w experiments to enable this.

PR #172, opened 2014-07-16
Idleruns

corpaul:
experiments for 1H, 24H and 1 week idle runs (without systemtap)


whirm:
Could you enable both the memory profiler and the manhole? We also need to add a passwd file to make manhole work:

```
tribler::1:1::/tmp/:/bin/sh
```

corpaul:
I enabled the profiler and manhole and set the dump times for each experiment. the passwd file will be handled by the meliae pull request

experiments/tribler_idle_run/1day.conf
```diff
@@ -0,0 +1,25 @@
+experiment_name = "TriblerIdleRun24hr"
+workspace_dir = "./"
```
whirm: This is the default workspace value, you can remove config options with default values so they will change if we update gumby.

experiments/tribler_idle_run/1day.conf
```diff
@@ -0,0 +1,25 @@
...
+
+#Custom variable to control the experiment execution time in seconds
+TRIBLER_EXECUTION_TIME = 86400
+
+# memory profiler stuff
+PROFILE_MEMORY = True
+PROFILE_MEMORY_INTERVAL = 3600
+PROFILE_MEMORY_PID_MODULO = 0
+MANHOLE_ENABLE = True
+MANHOLE_PORT = 2323
```
whirm: the ports in each experiment should be different in case they get executed in parallel on the same machine

corpaul:
fixed

experiments/tribler_idle_run/1day.conf
```diff
@@ -0,0 +1,24 @@
...
+# Enable use of virtualenv when running on the local machine (will be always used on the DAS4)
+use_local_venv = False
+
+#Custom variable to control the experiment execution time in seconds
+TRIBLER_EXECUTION_TIME = 86400
+
+# memory profiler stuff
+PROFILE_MEMORY = True
+PROFILE_MEMORY_INTERVAL = 3600
+PROFILE_MEMORY_PID_MODULO = 0
```
whirm: This is also a default value, the rest looks great

corpaul:
oops, forgot about that one, fixed

PR #174, opened 2014-07-17
Allchanlocal

corpaul:
local allchannel experiment

### DIFF
--- a/scripts/post_process_dispersy_experiment.sh
+++ b/scripts/post_process_dispersy_experiment.sh
@@ -37,7 +37,7 @@
 
 # Code:
 
-set -xe
+set -e
 
 # Step 1: Look for non-empty stderr files and print its contents
 
--- a/scripts/post_process_dispersy_experiment.sh
+++ b/scripts/post_process_dispersy_experiment.sh
@@ -53,7 +53,16 @@ if [ -z "$DISPERSY_STATISTICS_EXTRACTION_CMD" ]; then
     DISPERSY_STATISTICS_EXTRACTION_CMD=extract_dispersy_statistics.py
 fi
 
-cd $OUTPUT_DIR
+# @CONF LOCAL_OUTPUT_DIR: Output dir for local running experiments (so not on DAS4).
+# This is hack to avoid rewriting a bunch of scripts, as the scripts look for the directory structure $LOCAL_OUTPUT_DIR/localhost/localhost
+# For local experiments make sure to also set $OUTPUT_DIR
+# TODO fix this so that we don't need this variable
+if [ -n "$LOCAL_OUTPUT_DIR" ]; then
+	cd $LOCAL_OUTPUT_DIR
+else
+	cd $OUTPUT_DIR
+fi
+
 #Step 2: Extract the data needed for the graphs from the experiment log file.
 
 TEMPFILE=$(mktemp)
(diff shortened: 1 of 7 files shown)


scripts/post_process_dispersy_experiment.sh
```diff
@@ -53,7 +53,15 @@ if [ -z "$DISPERSY_STATISTICS_EXTRACTION_CMD" ]; then
...
-cd $OUTPUT_DIR
+# @CONF LOCAL_OUTPUT_DIR: Output dir for local running experiments (so not on DAS4).
+# This is hack to avoid rewriting a bunch of scripts, as the scripts look for the directory structure $LOCAL_OUTPUT_DIR/localhost/localhost
+# For local experiments make sure to also set $OUTPUT_DIR
+if [ -n "$LOCAL_OUTPUT_DIR" ]; then
+	cd $LOCAL_OUTPUT_DIR
+else
+	cd $OUTPUT_DIR
+fi
+
```
whirm, line 65: we could avoid this by doing

```
local_instance_cmd = "process_guard.py -c allchannel_client.py -c allchannel_client.py -c allchannel_client.py -c allchannel_client.py -c allchannel_client.py -t $EXPERIMENT_TIME -m $OUTPUT_DIR/local/local -o $OUTPUT_DIR "
```

instead

This way we don't need a new config option (`local_output_dir`) and its logic.

Other than that, it looks fine.

corpaul, line 65: That won't work because the nodes are still writing to $OUTPUT_DIR/1, $OUTPUT_DIR/2 etc.

corpaul:
I've looked into why processes are left hanging, but it seems that the process_guard does not capture SIGINT signals? Are they maybe catched and handled earlier by another script? I can't find it :(

corpaul:
SIGINT was captured, but only processes in the same process group were closed. I added a check for processes that have the gumby process as parent and close those as well. Seems to work locally and on Jenkins, also when aborting a job. So ready to merge

scripts/post_process_dispersy_experiment.sh
whirm, line 65: how would that be a problem? right now when we run it on the das4, we get a series of headnode/slavenode/processnumber so, if me make process guard write to local/local/ we will get the same structure. Am I missing smth?

corpaul, line 65: Some gets written to the -m directory and some gets written to the -o directory.. I tested it and it doesnt work

whirm:
That's weird, any process spawned by process guard should be in the process group. How can this process be out of the group?

corpaul:
Hmm that's weird.. when I check with ps the processes are indeed part of the process_guard group. I think I see what went wrong in my thinking process, I was fixing the check in run.py (which indeed has a different pid than the process_guard pgid). I'll check now why the process isn't closed by the process guard itself.

corpaul:
can you come on IRC? I have some questions

scripts/post_process_dispersy_experiment.sh
whirm, line 65: oh, I see, dispersyclient.py is also using OUTPUT_DIR... Ok, let's leave it as it is for now, could you add a TODO entry to solve this? I think the best would be for set a CWD for each subprocess so they don't have to manage this kind of stuff.

synctext:
thnx for those Gumby cleanups/fixes!

corpaul:
cleaned up the PR - it also contains a temp fix for cleaning up the processes that may be left hanging by process guard

PR #176, opened 2014-07-23
Don't call path.exists if output_dir is None

egbertbouman:
The Tribler pull requester is currently broken because os.path gets called with None as an argument (see http://jenkins.tribler.org/job/GH_Tribler_pull-request-tester_devel/2447/console).


whirm:
Hm, output_dir should never be None... @corpaul I think it has something to do with the local output dir changes?

whirm:
@corpaul forget it, wrong output_dir

PR #177, opened 2014-07-23
fix process guard

corpaul:
check if process group id is in our list of spawned process groups

### DIFF
--- a/scripts/process_guard.py
+++ b/scripts/process_guard.py
@@ -18,8 +18,11 @@ class ResourceMonitor(object):
 
         self.cmd_counter = 0
         self.pid_dict = {}
+
         self.files = []
         self.output_dir = output_dir
+
+        self.pgid_list = []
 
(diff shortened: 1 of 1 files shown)


scripts/process_guard.py
```diff
@@ -115,7 +115,7 @@ def update_pid_tree(self):
             io_file = path.join(pid_dir, 'io')
             if access(stat_file, R_OK) and access(io_file, R_OK):
                 pgrp = int(open(stat_file, 'r').read().split()[4])  # 4 is PGRP
-                if pgrp == self.process_group_id:
+                if pgrp in self.pgid_list:
```
whirm, line 119: pgid_list is always empty

corpaul, line 119: oops i mixed up pgid_list and pgid_dict.. fixed and cleaned up PR

can you check if this works for you? i couldn't reproduce your issue at my PC so difficult to test

PR #180, opened 2014-07-30
generate graphs for 1hr idle run

corpaul:
This is a conf file for the 1H idle run experiment with systemtap enabled. Also generates the process guard graphs.


experiments/tribler_idle_run/generate_stap_process_guard_graphs.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+stap_make_io_writes_report.sh $OUTPUT_DIR/report $OUTPUT_DIR/stap.csv "1H run report" 
+graph_process_guard_data.sh
```
whirm: small optimization: add & to both and then wait at the end of the script.

corpaul:
fixed

corpaul:
ready to merge

PR #182, opened 2014-07-31
fix buildswift script

corpaul:
now also accepts 'false' values :D

### DIFF
--- a/scripts/buildswift.sh
+++ b/scripts/buildswift.sh
@@ -40,9 +40,9 @@
 
 
 build () {
-    # @CONF_OPTION DEBUG_SWIFT: Set to any value if you want to enable Swift's debug output. (default is disabled)
-    if [ "${DEBUG_SWIFT,,}" == "false" ]; then
-        #Disable debug output
+    # @CONF_OPTION DEBUG_SWIFT: Set to true if you want to enable Swift's debug output. (default is disabled)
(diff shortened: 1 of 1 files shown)


scripts/buildswift.sh
```diff
@@ -41,7 +41,7 @@
 
 build () {
     # @CONF_OPTION DEBUG_SWIFT: Set to any value if you want to enable Swift's debug output. (default is disabled)
-    if [ -z "$DEBUG_SWIFT" ]; then
+    if [ "$DEBUG_SWIFT{,,}" == "false" ]; then
```
whirm: ${DEBUG_SWIFT,,}

scripts/buildswift.sh
```diff
@@ -50,7 +50,7 @@ build () {
 
 }
 
-if [ ! -z "$BUILD_SWIFT" ]; then
+if [ "$BUILD_SWIFT{,,}" == "true" ]; then
```
whirm: ditto

scripts/buildswift.sh
```diff
@@ -41,7 +41,7 @@
 
 build () {
     # @CONF_OPTION DEBUG_SWIFT: Set to any value if you want to enable Swift's debug output. (default is disabled)
```
whirm: Update the command line description and don't change the default behaviour.

corpaul:
ready to merge

PR #186, opened 2014-11-05
Seems that $TMPDIR is not created automatically on the DAS4 slave nodes?

whirm:
Also, cleanup fake $HOME after running.


whirm:
Do not merge yet

PR #196, opened 2015-02-16
Add dependencies

LipuFei:
Forgot to add `libsodium`, `python-cryptography`, and some other dependencies.


LipuFei:
@whirm ready

PR #207, opened 2015-04-09
Barterclient: fixed unused candidate id

### LINKED ISSUE #206: candidate_id not used in barter_clientclient.request_stat
@corpaul  showed me the method `request_stat` as an example of how to generate a single candidate in gumby for my own tests. But this method does not use a single candidate. Instead the method sends a request stat to every candidate.

I don't know what the wanted behavior is, but the parameter `candidate_id` is never used and it is not possible at the moment to sent a single request stat to a candidate using this method. It would be easily fixed by using the `candidate_id` as a key on `self.all_vars.`. The corresponding value will allow creation of the candidate in the same way as `c` is used for creation.

snorberhuis:
Fixes #206

### DIFF
--- a/experiments/dispersy/barter_client.py
+++ b/experiments/dispersy/barter_client.py
@@ -72,10 +72,21 @@ class BarterClient(AllChannelClient):
         self.scenario_runner.register(self.close, 'close')
 
     def request_stats(self, candidate_id=0):
+        """
+        Requests statistics from every candidate or a specific candidate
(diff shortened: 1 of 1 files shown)


corpaul:
@snorberhuis did you try to run this locally for both cases?

snorberhuis:
I did run it locally using barter_client for both cases. But I am unfamiliar with the results and could not really verify that way if it works. So I ran the same code in my own client and it worked there as intended.

corpaul:
ok, thanks for the fix!

PR #210, opened 2015-04-24
Gumby experiments for hidden services and tunnels with a working Tribler Session

rjruigrok:
Various fixes and improvements:
- Tunnel experiment fixed to work with new crypto
- Tunnel experiment not limited to DAS4 only
- TriblerDispersyExperimentClient added for experiments that need a Tribler session
- Hidden services experiment (using the Tribler session)


whirm:
Great, thanks a lot!

rjruigrok:
My quickest merged PR ever! :smile:

whirm:
haha

rjruigrok:
@whirm  When I run experiments locally, it finds candidates for the tunnel community very fast. But when I run the same experiment on Jenkins, it takes a very long time before it finds (or sometimes even does not find any) candidates. 

See for example the err-files in http://jenkins.tribler.org/job/Experiment_Tunnel_isolated_gorillas/ws/output/localhost/localhost/ which contain the number of verified dispersy candidates on different moments of time (around 20 seconds between each poke). 

Do you have an idea what might be the cause of this?

whirm:
I would enable logging for the walker and the tracker to see what's up with the peer discovery.

Are you running exactly the same experiment?

rjruigrok:
It's exactly the same experiment, same scenario, same config. I've enabled the logging, see results http://jenkins.tribler.org/job/Experiment_Tunnel_isolated_gorillas/174

From all the output it's not really clear to me what to look fo, do you see something strange in it?

whirm:
I see it's connecting to the external trackers, you should isolate the experiment, you can set the IP the tracker should listen to to 127.0.0.1 and the dispersy sessions should only connect to it too (I haven't checked the logs for that)

PR #217, opened 2015-06-05
Parallel tribler tests

whirm:
This introduces the side effect that the process guard graphs get really confusing as all the test annotations step on each other due to concurrency.

This gets us from ~38 to ~16 minutes for the whole Tribler test suite. If the tunnel tests got improved, this time could be easily halved.

Note that parallelisation is optional, by default we keep the old behaviour.


whirm:
@LipuFei @NielsZeilemaker ready to merge

PR #240, opened 2015-12-23
READY: MultiChain Experiments

Captain-Coder:
Since @snorberhuis is no longer active on his PR, I will take over his work. Please close his pr and review this one. (Or accept his and thus simplify this one). Accepting this PR should be synchronized with accepting Tribler/tribler#1784 eventually.

(Ignore the previous pull request, I borked my repo)

### DIFF
--- a/scripts/process_guard.py
+++ b/scripts/process_guard.py
@@ -12,7 +12,7 @@ from signal import SIGKILL, SIGTERM, signal
 from subprocess import Popen
 from time import sleep, time
 
-from psutil import Process
+from psutil import Process, AccessDenied
 
 OK_EXIT_CODE = 0
 TIMEOUT_EXIT_CODE = 3
@@ -319,12 +319,15 @@ class ProcessMonitor(object):
 
             if self.fd_file:
                 for child_process in p_children:
-                    if hasattr(self.psutil_process, 'num_fds'):
-                        self.fd_file.write("%.1f %s %d\n" %
-                                           (r_timestamp, child_process.pid, child_process.num_fds()))
-                    else:
-                        self.fd_file.write("%.1f %s %d\n" %
-                                           (r_timestamp, child_process.pid, child_process.get_num_fds()))
+                    try:
+                        if hasattr(self.psutil_process, 'num_fds'):
+                            self.fd_file.write("%.1f %s %d\n" %
+                                               (r_timestamp, child_process.pid, child_process.num_fds()))
+                        else:
+                            self.fd_file.write("%.1f %s %d\n" %
+                                               (r_timestamp, child_process.pid, child_process.get_num_fds()))
+                    except AccessDenied:
+                        pass  # Just ignore the file descriptors of this forbidden process
 
             if self.end_time and timestamp > self.end_time:  # if self.end_time == 0 the time out is disabled.
                 print "Time out, killing monitored processes."


whirm:
Don't merge back from devel, do a rebase instead

Captain-Coder:
Ok I rebased and neatified the changes, anything else that needs changing?

whirm:
I'll have a look at it ASAP

experiments/multichain/multichain_client.py
```diff
@@ -0,0 +1,177 @@
...
+        self.scenario_runner.register(self.increase_kbytes_received, 'increase_kbytes_received')
+        self.scenario_runner.register(self.schedule_block, 'schedule_block')
+        self.scenario_runner.register(self.close, 'close')
+
+    def set_multichain_type(self, multichain_type='MultiChainGumbyCommunity'):
+        """
+        Sets the community class of this gumby node to a special community.
+        """
+        msg("CommunityType: %s" % multichain_type)
+        if multichain_type == 'MultiChainDelayCommunity':
```
whirm: It'd be nicer to `import Tribler.community.multichain.community as multichain_community` and do a hasattr()/getattr() instead of daisy-chaining all those if's.

experiments/multichain/multichain_client.py
```diff
@@ -0,0 +1,177 @@
...
+            msg("Starting MultiChain client with: " + MultiChainNoResponseCommunity.__name__)
+            self.community_class = MultiChainNoResponseCommunity
+        elif multichain_type == 'MultiChainCommunityCrawler':
+            msg("Starting MultiChain client with: " + MultiChainGumbyCommunityCrawler.__name__)
+            self.community_class = MultiChainGumbyCommunityCrawler
+        else:
+            raise RuntimeError("Tried to set to unknown community:%s." % multichain_type)
+
+    def online(self, dont_empty=False):
+        DispersyExperimentScriptClient.online(self, dont_empty)
```
whirm: use `super`

experiments/multichain/multichain_crawl_simple.conf
```diff
@@ -0,0 +1,66 @@
+# This is a gumby experiment config file
```
whirm: Cleanup the file to contain only the stuff relevant to this experiment.

experiments/multichain/multichain_integrated_client.py
```diff
@@ -0,0 +1,123 @@
...
+        """ Integrated callbacks"""
+        self.scenario_runner.register(self.start_multichain, 'start_multichain')
+        self.scenario_runner.register(self.start_scheduler, 'start_scheduler')
+
+    def set_multichain_type(self, multichain_type='MultiChainCommunity'):
+        """
+        Sets the community class of this gumby node to a special community.
+        """
+        msg("CommunityType: %s" % multichain_type)
+        if multichain_type == 'MultiChainDelayCommunity':
```
whirm: What about having a common superclass, this class has duplicated code with MultichainClient.

experiments/multichain/processing/DatabaseReader.py
```diff
@@ -0,0 +1,260 @@
...
+
+
+class DatabaseReader(object):
+
+    def __init__(self, working_directory):
+        """
+        Parent class that contains functionality for both gumby and single database readers.
+        DatabaseReaders process experimentation information.
+        """
+        """ Either contains all information all ready or all information will be aggregated here."""
```
whirm: String comments?

experiments/multichain/processing/DatabaseReader.py
```diff
@@ -0,0 +1,260 @@
...
+    def add_block_to_graph(self, block_id, block):
+        # The block.id is not usable because it differs from the actual hash as the PKs are missing.
+        block_id_encoded = base64.encodestring(block_id)
+        # Add the block to the graph
+        self.graph.add_node(block_id_encoded)
+        # Add the edges of the blocks
+        self.graph.add_edge(block_id_encoded, base64.encodestring(block.previous_hash_requester))
+        self.graph.add_edge(block_id_encoded, base64.encodestring(block.previous_hash_responder))
+        if block.previous_hash_requester == block.previous_hash_responder:
+            """ Follow up block"""
```
whirm: here too

experiments/multichain/processing/DatabaseReader.py
```diff
@@ -0,0 +1,260 @@
...
+    def get_database(self, db_path):
+        return MultiChainExperimentAnalysisDatabase(self.MockDispersy(), db_path)
+
+
+def string_is_int(string):
+    try:
+        int(string)
+        return True
+    except ValueError:
+        return False
```
whirm: no newline at end of file (there's several of them, fix them all please)

experiments/multichain/processing/main-gumby-integrated.py
```diff
@@ -0,0 +1,25 @@
+"""
+Performs the experimentation processing of a gumby experiment.
```
whirm: experimentation processing?

gumby/experiments/TriblerDispersyClient.py
```diff
@@ -59,7 +59,7 @@ def _do_start():
 
             logging.error("Tribler Session started")
             self.annotate("Tribler Session started")
-
+
```
whirm: Remove this and make sure you remove all trailing white spaces from all files you create in this PR.

experiments/multichain/multichain_crawl_simple.conf
```diff
@@ -0,0 +1,66 @@
...
+#
+# Command used to start (the local part of) the experiment. This will be executed concurrently with its remote counterpart if set (see below)
+local_instance_cmd = "process_guard.py -c multichain_client.py  -c multichain_client.py -c multichain_client.py -c multichain_client.py  -c multichain_client.py -c multichain_client.py"
+
+#
+# Command used to start (the remote part of) the experiment. This will be executed concurrently with its local counterpart if set (see above)
+# remote_instance_cmd =
+#
+# Command used to post-process the data obtained from the experiment (aggregate data, graph stuff, etc.), this command will be run locally.
+# post_process_cmd =
```
whirm: Is this experiment really not post-processing/generating any graphs?

experiments/multichain/multichain_integrated.conf
```diff
@@ -0,0 +1,32 @@
+# Gumby experiment with the anonymous download integrated with the MultiChain.
```
whirm: This one neither?

whirm:
The code looks mostly Ok, needs some cosmetic/style stuff and a couple of minor refactors.

Captain-Coder:
Picked the low haning fruit. Nice onelines with sed got most of the text edits in one go :P 

There is indeed quite a bit of overlap between multichain_client and multichain_integrated_client. But their base classes have different uses, one only starts dispersy, the other starts pretty much the whole tribler core. This difference goes all the way back to gumby's dispersyclient.py and TriblerDispersyClient.py both being separate experiment servers. The most straightforward way of uniting the multichain clients would be to [make the super dynamic](http://stackoverflow.com/questions/17599850/dynamically-choosing-class-to-inherit-from) but this makes my brain hurt. Atleast, i'm not sure about the practical/longer term implications of doing this, it just feels wrong.

About experiments not having post processing: there is obviously code for this, and i don't know why @snorberhuis never used the post_processing config. I'll ask him, it might just be manual commandline invocations though.

experiments/multichain/multichain_client.py
Captain-Coder: Yes importing with a specific name might be usefull in this case, it discriminates against other communities.

Also the derivatives, DelaySignCommunity and DenySignCommunity don't have different attributes, so I don't see how a hasattr() can eliminate the if chain? I can rewrite this to be a bit better/generic without one though. What is it about the if chaining that irks you btw?

snorberhuis:
I just added the post processing to my bash startup script. I think I did try to use the config, but it didn´t work because of python path issues.

snorberhuis:
I also tried to have a nice inheritance structure to remove duplicate code in the two clients, but this failed, because of the problem @Captain-Coder said. The best approach might be to refactor to common functions not in a class and use these. 

The differences in both superclasses makes this hard, but maybe you can refactor away these differences first and make it easier.

whirm:
Haven't looked closely at the code, but can't this be solved with multiple inheritance?

Captain-Coder:
Uhhmm... you recommended to use super().

I have since discovered there are two styles of classes in python... the base class that the experiment server is based on (LineReciever from twisted) is an old-style class. Calling super() on this fails. Hence all the old style direct calling of base class initializers. The base DispersyClient, HiddenServicesClient and even ExperimentClient all use this style (obviously).

Also @synctext suggest _not_ using multiple inheritance, I tend to agree with him. As alternative, perhaps we can refactor atleast the multichain part to a sort of factory pattern that "Makes a client", the rest of the code should not care if this is a dispersy or tribler core client.

synctext:
@whirm The multichain stuff can be included and enabled in the upcoming release I hope. With 2 students it can be ready next week possibly.

whirm:
@Captain-Coder super() works with new style classes (follows the inheritance chain) but not with old style ones (which means that you have to explicitly call the parent class method).

Buuut, I thought I had converted the experiment classes to new style already but turns out I didn't (trivial thing to do, already merged), so my bad. Please do a rebase.

I usually try to avoid multiple inheritance but (at a first glance) this is a pretty straightforward case, you need a class to hold the methods that are common across two semi-related classes. I would go that way as it requires minimal refactoring and it's still easy to understand what's going on.

I'm refactoring all the other experiment classes now to use `super()` so nobody gets bitten in the bottom by this in the future.

I'll answer to the other comments shortly.

whirm:
@synctext I don't think it's a good idea to introduce new features at this stage of stabilizing, look at how long it's taking us to release 6.5 as it is. 

Let's merge it to `devel` and branch off for stabilization right after 6.5 is out.

experiments/multichain/multichain_client.py
whirm: What I mean is that it's more generic and concise to do something like this:

``` python
from Tribler.community.multichain import community as multichain_communities

[...]

def set_multichain_type(self, multichain_type='MultiChainGumbyCommunity'):
    self.community_class = getattr(multichain_communities, multichain_type, None)
    if not self.community_class:
        raise RuntimeError("Tried to set to unknown community:%s." % multichain_type))
    msg("Starting MultiChain client with :  %r", self.community_class)
```

experiments/dispersy/hiddenservices_client.py
```diff
@@ -69,19 +72,19 @@ def init_community(self, become_exitnode=None, no_crypto=None):
...
         if not self.security_limiters:
             tunnel_settings.max_traffic = 1024 * 1024 * 1024 * 1024
 
-        tunnel_settings.min_circuits = 3
-        tunnel_settings.max_circuits = 5
+        tunnel_settings.min_circuits = self.min_circuits
+        tunnel_settings.max_circuits = self.max_circuits
 
-        logging.error("My wan address is %s" % repr(self._dispersy._wan_address[0]))
+        logging.debug("My wan address is %s" % repr(self._dispersy._wan_address[0]))
```
whirm: I guess you forgot that one?

whirm:
What's the status of this?

whirm:
ok to test

whirm:
Let me know when this is ready

Captain-Coder:
Ready to merge as far as i can tell. However, I am not able to run all experiments. So someone with access should probably run the multichain (and tunnel) experiements on BBQ.

whirm:
I'll wait for that to happen then

whirm:
Maybe that should be worked on guys?

Captain-Coder:
@whirm and @lfdversluis , This has been pending for quite some time. Please review and merge.

Captain-Coder:
@lfdversluis I reordered, merged and squashed commits to make it all logical. Can you see if it's ok like this?

devos50:
@Captain-Coder I suppose you wanted to tag me? :P

experiments/__init__.py
```diff
@@ -0,0 +1 @@
+__author__ = 'norberhuis'
```
devos50: Please remove these author tags and replace them by a meaningful description what the package contains.

experiments/multichain/__init__.py
```diff
@@ -0,0 +1 @@
+__author__ = 'norberhuis'
```
devos50: Same as comment above.

experiments/multichain/multichain_client.py
```diff
@@ -0,0 +1,135 @@
...
+        if community_type == 'MultiChainDelayCommunity':
+            msg("Starting MultiChain client with: " + MultiChainDelayCommunity.__name__)
+            self.community_class = MultiChainCommunity
+        elif community_type == 'MultiChainNoResponseCommunity':
+            msg("Starting MultiChain client with: " + MultiChainNoResponseCommunity.__name__)
+            self.community_class = MultiChainNoResponseCommunity
+        else:
+            raise RuntimeError("Tried to set to unknown community.")
+
+    def online(self):
```
devos50: Is this override necessary?

experiments/multichain/multichain_client.py
```diff
@@ -0,0 +1,135 @@
...
+        msg("%s: Requesting Signature for candidate: %s" % (self.my_id, candidate_id))
+        if candidate_id == 0:
+            for c in self.all_vars.itervalues():
+                candidate = self._community.get_candidate((str(c['host']), c['port']))
+                print("Member: %s" % candidate.get_member())
+                self._community.publish_signature_request_message(candidate)
+        else:
+            target = self.all_vars[candidate_id]
+            candidate = self._community.get_candidate((str(target['host']), target['port']))
+            print("Candidate: %s" % candidate.get_member())
```
devos50: Please don't use print statements but use the logger instead (there are several print statements in this file).

experiments/multichain/multichain_client.py
```diff
@@ -0,0 +1,135 @@
...
+        self._community.unload_community()
+
+
+class MultiChainDelayCommunity(MultiChainCommunity):
+    """
+    Test Community that delays signature requests.
+    """
+    delay = 3
+
+    def __init__(self, *args, **kwargs):
```
devos50: Unnecessary init?

experiments/multichain/multichain_client.py
```diff
@@ -0,0 +1,135 @@
...
+        super(MultiChainCommunity, self).__init__(*args, **kwargs)
+
+    def allow_signature_request(self, message):
+        """
+        Ignore the signature requests.
+        :param message: the to be delayed request
+        """
+        self._logger.info("Received signature request that will delayed for %s." % self.delay)
+        sleep(self.delay)
```
devos50: Please don't use `sleep` statements. Use deferreds instead.

experiments/multichain/multichain_client.py
```diff
@@ -0,0 +1,135 @@
...
+        self._logger.info("Delay over.")
+        super(MultiChainCommunity, self).allow_signature_request(message)
+
+
+class MultiChainNoResponseCommunity(MultiChainCommunity):
+    """
+    Test Community that does not respond to signature requests.
+    """
+
+    def __init__(self, *args, **kwargs):
```
devos50: Unnecessary init

experiments/multichain/multichain_client.py
```diff
@@ -0,0 +1,135 @@
...
+        super(MultiChainNoResponseCommunity, self).__init__(*args, **kwargs)
+
+    def allow_signature_request(self, message):
+        """
+        Ignore the signature requests.
+        :param message: the to be ignored request
+        """
+        self._logger.info("Received signature request that will be ignored.")
+        return
```
devos50: Please remove this return

experiments/__init__.py
```diff
@@ -1 +0,0 @@
-__author__ = 'norberhuis'
```
devos50: Please don't edit stuff you introduced in previous commits.

experiments/multichain/multichain_client.py
```diff
@@ -11,7 +11,7 @@
 pythonpath.append(path.abspath(path.join(path.dirname(__file__), '..', '..', '..', "./tribler")))
 
 from Tribler.dispersy.candidate import Candidate
-from Tribler.community.multichain.community import MultiChainCommunity
+from Tribler.community.multichain.community import MultiChainCommunity, MultiChainCommunityCrawler
```
devos50: You're changing/fixing stuff here that you introduced in a previous commit. I suggest that you have one commit to create the `multichain_client.py` file. Please don't split this across several commits.

experiments/multichain/multichain_crawler.scenario
```diff
@@ -1,16 +1,9 @@
 @0:0 set_master_member 3081a7301006072a8648ce3d020106052b81040027038192000406712a1e5381d030ae25697bd6626810af48ed5094e5faae18a8b91e482e60b8681b26bf6aca786ad81b207ffb5a755956b4d6353c9dc487eb1bfdd17b66bd0de444e01a375fcf3401177a7bb102da7c5f50ad9375fe4a1ae1baca9d47870f56c841169fc36c2b113bafbadacab9f2b3c7773b3d82a54cda8fbde41b9e0e571ba631997a8259cc78e7f262e933542b51
-@0:0 set_community_class MultiChainDelayCommunity {2}
-@0:0 set_community_class MultiChainNoResponseCommunity {4}
+@0:0 set_multichain_type MultiChainCommunityCrawler {1}
```
devos50: The same applies here: just make sure you have one commit that introduces this `.scenario` file. Don't change it in subsequent commits.

experiments/multichain/multichain_integrated_client.py
devos50: I agree with this, creating a superclass that can be reused saves much code.

gumby/experiments/TriblerDispersyClient.py
devos50: This hasn't been addressed yet.

experiments/multichain/processing/DatabaseReader.py
devos50: I noticed this also in other commits. In general, we are using comments starting with # (hashtag) if it's only one line. Please change this so it follows our style.

experiments/multichain/processing/__init__.py
```diff
@@ -0,0 +1 @@
+__author__ = 'norberhuis'
```
devos50: Author tag again, see my first comment about author tags.

experiments/multichain/__init__.py
```diff
@@ -0,0 +1 @@
+
```
devos50: Please replace this change by a description of the package content/purpose.

experiments/multichain/r/total_data_in_blocks.R
```diff
@@ -24,4 +24,4 @@ if (file.exists("output/mids_data_down.dat")) {
 
     # Turn off device driver (to flush output to PDF)
     dev.off()
-}
\ No newline at end of file
+}
```
devos50: Please move this change to the commit where you introduced this file.

gumby/experiments/TriblerDispersyClient.py
```diff
@@ -59,7 +59,7 @@ def _do_start():
 
             logging.error("Tribler Session started")
             self.annotate("Tribler Session started")
-    
+
```
devos50: Please move this change to the commit where you introduced this file.

experiments/multichain/multichain.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python
```
devos50: See https://github.com/Tribler/gumby/pull/238

experiments/multichain/multichain.py
```diff
@@ -79,6 +83,33 @@ def introduce_candidates(self):
...
+        """
+        msg("%s: Increasing bytes received: %s For candidate: %s" % (self.my_id, amount_received, candidate_id))
+        target = self.all_vars[candidate_id]
+        peer = (str(target['host']), target['port'])
+        self.multichain_scheduler.update_amount_received(peer, int(amount_received)*1000)
+
+    def schedule_block(self, candidate_id):
+        """
+        Finish up the download.
+        :param candidate_id:
```
devos50: Please explain candidate_or leave this parameter our of the comments ;)

devos50 (commented):
I still see too many files being modified in subsequent commits which makes it harder for me to review this PR. Again, please make sure that you are not creating files in one commit, only to modify them in following commits.

When this is fixed, I will continue the review.

Captain-Coder:
@devos50 yes sorry meant to tag you instead of Laurens.

"Please don't edit stuff you introduced in previous commits" <- this is a problematic way of thinking i.m.o.

I'm not sure how you expect any (feature) PR to pass that requirement. Unless its a microscopic change. I'll say upfront that I'm not a fan of rebasing to begin with, since it is nothing short of geschiedvervalsing. But ok it's a requirement of this project so let's step onto this slippery slope. Removing the fixup commits, ok that removes some clutter, I can get behind that. Reordering commits to make a consistent story. Meh, but ok. Moving file mode changes to coincide with file creation, removes more noise. Rewording some commits, ok. But you are now suggesting that I rewrite all commits such that each file is touched/created only once in their perfect end state? Ehh, no, that is sliding off the slippery slope full orwellian 1984 style. There is not much point in keeping a history if you're only gonna commit perfect final stuff. It removes the evolution that the feature went through. So I'll squash it to one big commit if you like, but I won't artificially create a history that has zero relation to it's actual history.

Say I have some project on which I start a feature branch. I create file A.py, that has a first simple iteration of the feature. Seems like a good place to commit since it should atomically introduce the feature. Both before and after the commit the project should work, the only difference being the new feature. Then I realize that it could use a little efficiency/polishing/logging/whatever-other-improvement and I add file B.py. But I have to modify A.py to integrate it with B.py. So I commit the changed A.py and the new B.py. Before commit 1, in between commit 1 and 2 and after commit 2 there is a consistent state of the project and this history shows the evolution of the feature. Cheers all around. However, if I reorder stuff the way you are suggesting I'd have a first commit with A.py that integrates with B.py, but no B.py yet, cause that is a commit later. And the original unsophisticated version of A.py is lost. Only after the last commit is the story consistent again. What's the point of separating the commits then?

So the end result of your "it must fit my commit-by-commit reviewing style" is to just squash in one big commit. I've got to spend my time (which I don't have too much of) doing thesis work, not nitpicking the commit history that has no effect on the end result and removes all value from said history.

devos50:
Ok, maybe I formulated this statement too broad. Let me give an example in this PR: in https://github.com/Tribler/gumby/pull/240/commits/7ee04d47178f902393343de27896b472ce6f34cc, the `__init__.py` files were introduced which contain an author tag. In https://github.com/Tribler/gumby/pull/240/commits/0fefbbea0cbc02ed21bff11559304bc2d0235685, these tags are removed. This change should be squashed since it is a 'fix' for a problem introduced in earlier commits.

Another example: the sleep statement got removed from `multichain_client.py` in https://github.com/Tribler/gumby/pull/240/commits/a8ece02aa5ec89f7552e92f2d9ff46f3d110da49 and has been replaced by a `callLater` statement. Again, this is a fix for something that was introduced in an earlier commit and should be squashed.

Also, regarding the configuration files (`.conf`) and scenario files (`.scenario`), they are modified in several commits. For instance, various configuration files got pushed with many unnecessary configuration options commented out. These commented-out options were removed in subsequent commits. They should be squashed since having this change in the commit history adds no value.

In general, try to squash fixes you made to your own code so these are not visible in the commit history. This makes the commit history cleaner and makes it easier to use the blame functionality in git.

It's not that much work if you use a proper git editor (there are many vim/emacs plugins and GUI tools available which makes it easy to use the hunk functionality of git and move parts of a commit around.

synctext:
@pimveldhuisen OK, excellent. Please clean & fix this code.

devos50:
@pimveldhuisen please be aware that the `stop` method of Dispersy returns a Deferred (https://github.com/Tribler/dispersy/blob/devel/dispersy.py#L2229). I'm not sure whether Gumby accounts for that but in Tribler, we didn't. So please wait for this deferred to fire before you continue doing other things in Gumby 👍

PR #241, opened 2015-12-29
Fix for UCS when building virtual env

Captain-Coder:
I ran in to this when running various random gumby experiments. It would compile c-python with the default UCS2, while the system libraries of newer distributions are compiled with UCS4 as default. Everything that is packaged into the virtual env is thus not compatible with the python binary that is packaged along.

### DIFF
--- a/scripts/build_virtualenv.sh
+++ b/scripts/build_virtualenv.sh
@@ -40,7 +40,7 @@
 #
 
 # Increase this every time the file gets modified.
-SCRIPT_VERSION=17
+SCRIPT_VERSION=18
 
 # Code:
 set -e
@@ -149,6 +149,9 @@ if [ ! -e $VENV/inst/.completed.$SCRIPT_VERSION ]; then
     if [ ! -e $VENV/inst/bin/python ]; then
         pushd cpython-2011
         hg checkout dtrace-issue13405_2.7
+        if ! ( python -c 'import sys; print(sys.maxunicode)' | grep -q '65535' ); then
+            EXTRA_CONFIG_OPTS="$EXTRA_CONFIG_OPTS --enable-unicode=ucs4"
+        fi
         LDFLAGS="-Wl,-rpath=$VENV/inst/lib" ./configure $EXTRA_CONFIG_OPTS --prefix=$VENV/inst --enable-shared
         cp Modules/Setup.dist Modules/Setup
         make -j$(grep process /proc/cpuinfo | wc -l)


scripts/build_virtualenv.sh
```diff
@@ -149,7 +149,11 @@ if [ ! -e $VENV/inst/.completed.$SCRIPT_VERSION ]; then
     if [ ! -e $VENV/inst/bin/python ]; then
         pushd cpython-2011
         hg checkout dtrace-issue13405_2.7
-        LDFLAGS="-Wl,-rpath=$VENV/inst/lib" ./configure $EXTRA_CONFIG_OPTS --prefix=$VENV/inst --enable-shared
+        if ( python -c 'import sys; print(sys.maxunicode)' | grep -q '65535' ); then
+            LDFLAGS="-Wl,-rpath=$VENV/inst/lib" ./configure $EXTRA_CONFIG_OPTS --prefix=$VENV/inst --enable-shared
+        else
+            LDFLAGS="-Wl,-rpath=$VENV/inst/lib" ./configure $EXTRA_CONFIG_OPTS --prefix=$VENV/inst --enable-shared --enable-unicode=ucs4
```
whirm: Just extend LDFLAGS in the `if` instead of copypasting the whole thing with an extra arg.

whirm:
Nice, thanks!

But the experiments shouldn't need on that venv if you aren't running them on the DAS4 (except for the ones relying on systemtap which need a patched python runtime)

scripts/build_virtualenv.sh
Captain-Coder: With this syntax the LDFLAGS are only set for this one invocation of ./configure. Also configure doesn't usually take its configuration options from LDFLAGS. And are you sure LDFLAGS is not used anywhere else in the script that would mess things up?

Captain-Coder:
Ok I changed it around a bit, it now works by extending EXTRA_CONFIG_OPTS, which is only used for this one invocation.

Also I know my situation doesn't require v-env, I just happend to run into this when setting up gumby. I ran a bunch of experiments to check I had gumby set up right and this came up.

scripts/build_virtualenv.sh
whirm: I don't know why I wrote LDFLAGS, sorry. I meant the (whole) command line, something like:

``` bash
if CONDITION then EXTRA_PYTHON_CONFIG_ARGS="asdfasdf"
fi
LDFLAGS=foo bar ./configure --whatever $EXTRA_PYTHON_CONFIG_ARGS
```

whirm:
It's amazing that this script still builds with that minor fix after all this time :)

PR #244, opened 2016-01-05
Changed lowercase check to support bash 3.x

devos50:
The case modifiers are not supported in bash 3.x (which is the version that OS X is using). In order to run the unit tests on OS X, I propose a different way of checking lowercase strings by overriding the `nocasematch` setting (thanks to the #bash IRC channel).

### DIFF
--- a/scripts/run_nosetests_for_jenkins.sh
+++ b/scripts/run_nosetests_for_jenkins.sh
@@ -39,6 +39,7 @@
 
 
 set -e
+shopt -s nocasematch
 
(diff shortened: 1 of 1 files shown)


whirm:
Just set it for the relevant block, let's not leave it set for the whole script.

(also, have you tested in on linux to make sure it still works fine there?)

devos50:
@whirm good point, will do that and I will test it on Linux in a moment.

devos50:
Confirmed to work on linux, also squashed the commits.

whirm:
Great!

PR #245, opened 2016-01-05
Added condition for OS X environment regarding sort

devos50:
The OS X built-in version of sort does not support the `-r` flag. `gsort` (installed with brew) has support for this flag.

I've tested both on OS X and Linux.

### DIFF
--- a/scripts/run_nosetests_for_jenkins.sh
+++ b/scripts/run_nosetests_for_jenkins.sh
@@ -96,8 +96,16 @@ else
         COUNT=0
         OLD_IFS=$IFS
         IFS=$'\n'
+
+        SORT_CMD="sort"
+        if [ $(uname) == "Darwin" ]; then
+            SORT_CMD="gsort"
+        fi
+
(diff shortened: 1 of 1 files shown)


synctext:
http://win-bash.sourceforge.net

bash.exe for windows without special .dll files.
Perhaps handy for windows port or cygwin.

scripts/run_nosetests_for_jenkins.sh
```diff
@@ -96,8 +96,15 @@ else
         COUNT=0
         OLD_IFS=$IFS
         IFS=$'\n'
+
+        if [ $(uname) == "Darwin" ]; then
+            LINES=$(find ${NOSE_TESTS_TO_RUN} -type f -iname "test_*.py" | gsort -R --random-source=/dev/zero | xargs -n${BUCKET_SIZE})
```
whirm: I would rather do an `SORT=gsort -R` or otherwise on the if and then that variable it in the pipe to avoid duplicating the whole line for that minor difference.

devos50: Consider it done :)

whirm:
Good, squash it an I'll merge it.

devos50:
@whirm ready for merge

PR #246, opened 2016-01-11
Don't use the wrap_in_vnc script in an OS X environment

devos50:
The wrap_in_vnc script is used to wrap the Tribler tests in a vnc environment on the headless Linux server. On OS X we will not use vnc and let Tribler run normally.

### DIFF
--- a/scripts/wrap_in_temp_home.sh
+++ b/scripts/wrap_in_temp_home.sh
@@ -1,0 +1,38 @@
+#!/bin/bash -ex
+
+# Looks like $TMPDIR doesn't exist in the DAS4
+if [ ! -z "$TMPDIR" ]; then
+    mkdir -p "$TMPDIR"
+fi
+
+export HOME=$(mktemp -d)
+
+if [ ! -e $HOME ]; then
+    echo "Something went wrong while creating the temporary execution dir, aborting."
+    exit 3
+fi
+
+echo "Using temporary home directory: $HOME"
+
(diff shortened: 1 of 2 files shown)


whirm:
@devos50 we still want to do the fake home thingie, maybe you want to split the wrap-in-vnc script in two and make the first one call the other.

devos50:
@whirm I'm done with this PR. I've wrapped some (Linux-specific) statements in an if-statement, requiring minimal changes. If you have any other idea or suggestion, please let me know.

whirm:
I don't think it makes sense to have a wrap_in_vnc script that doesn't actually wraps in vnc depending on the platform you run it in. Let's just split it in two scripts.

devos50:
That's true, I will split it in a moment.

devos50:
@whirm I've split the script in two parts. It's still working on Linux. The `wrap_in_vnc` script calls `create_temp_home.sh` which returns the name of the new created home directory.

whirm:
Better to have a wrap_in_temp_home.sh script that receives a command to run and cleans up after itself, like wrap_in_vnc.sh does, this way you can use either of them the same way and you don't have to clean up after it.

devos50:
@whirm I've created a `wrap_in_temp_home` script and tested it on Linux. The `wrap_in_temp_home` is called by `wrap_in_vnc`.

whirm:
Looks fine!

Squash it!

devos50:
@whirm done :+1:

PR #247, opened 2016-01-15
Added pylint checker script

devos50:
This scripts runs the pylint checker and has three input options:
-i : the input directory of pylint
-o : the output directory where pylint will store the `pylint.log` and `pylint.out` files
-e : an optional option with directories to exclude (comma-separated)

### DIFF
--- a/experiments/tribler/run_pylint_checker.conf
+++ b/experiments/tribler/run_pylint_checker.conf
@@ -1,0 +1,11 @@
+experiment_name = TriblerPylintChecker
+
+local_setup_cmd = tribler_experiment_setup.sh
+
+pylint_run_dir = tribler
+
+pylint_ignore_dirs = .git,dispersy,pymdht,libnacl
+
+local_instance_cmd = run_pylint_check.sh
+
+use_local_venv = False
--- a/scripts/run_pylint_check.sh
+++ b/scripts/run_pylint_check.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+# This scripts runs the pylint checker on a given input directory
+
+# @CONF_OPTION PYLINT_IGNORE_DIRS: Specify which directories should be ignored by Pylint (default is '')
+
+# @CONF_OPTION PYLINT_RUN_DIR: Specify from which directory pylint should run (default is $PWD)
+if [ ! -z "$PYLINT_RUN_DIR" ]; then
+cd $PYLINT_RUN_DIR
+fi
+
+echo "Running pylint (ignored directories: ${PYLINT_IGNORE_DIRS})"
+pylint --ignore=${PYLINT_IGNORE_DIRS} --output-format=parseable --reports=y Tribler > ${OUTPUT_DIR}/pylint.out 2> ${OUTPUT_DIR}/pylint.log


scripts/run_pylint_check.sh
```diff
@@ -0,0 +1,38 @@
...
+usage
+fi
+
+if [ -z "${e}" ]; then IGNORE=""; else IGNORE="--ignore=${e}"; fi
+
+echo "Running pylint with:"
+echo "Input directory: ${i}"
+echo "Output directory: ${o}"
+echo "Ignoring directories: ${e}"
+pylint ${IGNORE} --output-format=parseable --reports=y  ${i} > ${o}/pylint.out 2> ${o}/pylint.log
```
whirm: Better to use gumby-exported $OUTPUT_DIR var for the output, defaulting to PWD.
Remember to add some documentation with the correct comment syntax so it is found by the config generator. (https://github.com/Tribler/gumby/blob/devel/scripts/generate_config_file.py#L46)

In general it's better to use config options than command arguments as it will be easier to discover by other users.

whirm: Remember to add a newline at the end of the file

devos50: Ok, is it also better to use config options for the output directory and the excluded directories?

whirm: Yes, that would be the best, in general the idea is to try to use config vars for everything and document them so they are discoverable. It also makes scripts simpler (you don't have to do all this argument parsing stuff that you did).

devos50: Great, will fix it in a moment (this parsing of arguments in bash is really annoying to do indeed...)

devos50:
@whirm the script seems to work: https://jenkins.tribler.org/job/GH_Tribler_PR_pylint_check/44/console so I'm finished with this PR.

whirm:
@devos50 it should have default values for the output stuff (Just use $OUTPUT_DIR by default) so people don't have to add the same config options on all experiments.

whirm:
Look at other gumby scripts for a quick way to do that. add this and squash it and I'll merge it.

devos50:
@whirm should I just add this to the comment, so for instance:

```
# @CONF_OPTION PYLINT_IGNORE_DIRS: Specify which directories should be ignored by Pylint (default is $OUTPUT_DIR)
```

?

whirm:
You also need to set the default value for the variable, gumby won't do that for you.

whirm:
Or just use OUTPUT_DIR, I don't think anyone will need to customize where the pylint files are written ever.

devos50:
Is this correct?

```
if [ -z "$PYLINT_IGNORE_DIRS" ]; then
    PYLINT_IGNORE_DIRS=""
fi
```

(and the same for the other 2 variables)

whirm:
that checks if the variable it's _not_ a nil string and sets it to a nil string in that case.

devos50:
@whirm true, removed the exclamation mark now, typed it a little too fast lol. Is this the way to go?

whirm:
it doesn't really do much, "$VAR" will expand to "" both when $VAL is not set and when it is set to ""

devos50:
@whirm Ok, this example is bad, but it could be used for the `PYLINT_INPUT_DIR` and `PYLINT_OUTPUT_DIR` directory? Like:

```
if [ -z "$PYLINT_INPUT_DIR" ]; then
    PYLINT_INPUT_DIR=${OUTPUT_DIR}
fi
```

whirm:
yes, but I wouldn't bother, I would just use $OUTPUT_DIR in both places and be done with it as chances are that nobody will ever need set that to anything custom.

whirm:
perfect, squash and I'll merge it

experiments/tribler/run_pylint_checker.conf
```diff
@@ -0,0 +1,15 @@
...
+
+local_setup_cmd = tribler_experiment_setup.sh
+
+nose_run_dir = tribler
+
+nose_tests_to_run = Tribler/Test/
+
+local_instance_cmd = run_pylint_check.sh
+
+post_process_cmd = graph_process_guard_data.sh
```
whirm: We aren't running that on the process guard for now, so just remove that line.

experiments/tribler/run_pylint_checker.conf
```diff
@@ -0,0 +1,15 @@
...
+
+use_local_venv = False
+
+build_swift = True
```
whirm: We don't have swift on Tribler anymore, so remove that too.

experiments/tribler/run_pylint_checker.conf
```diff
@@ -0,0 +1,15 @@
+experiment_name = TriblerPylintChecker
+
+local_setup_cmd = tribler_experiment_setup.sh
+
+nose_run_dir = tribler
+
+nose_tests_to_run = Tribler/Test/
```
whirm: we aren't running nose on this experiment

experiments/tribler/run_pylint_checker.conf
```diff
@@ -0,0 +1,9 @@
+experiment_name = TriblerPylintChecker
+
+local_setup_cmd = tribler_experiment_setup.sh
+
+nose_run_dir = tribler
```
whirm: forgot this one :)

whirm: oh, sorry, nevermind

scripts/run_pylint_check.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+# This scripts runs the pylint checker on a given input directory
+
+# @CONF_OPTION PYLINT_IGNORE_DIRS: Specify which directories should be ignored by Pylint (default is '')
+
+# @CONF_OPTION NOSE_RUN_DIR: Specify from which directory nose should run (default is $PWD)
```
whirm: "*pylint should run"

whirm:
do the last modification and squash it, I'll merge it right away

devos50:
@whirm ready for merge.

PR #248, opened 2016-01-19
Create script to build venv with dependencies

wouterjsmit:
Installs all dependencies as they are listed in a file
"requirements.txt" in the experiment directory.


whirm:
Add some documentation so the config template generator finds the script and extract the relevant info please. (you can have a look at any of the scripts mentioned in the template config for the expected syntax)

wouterjsmit:
Added a config option and commentary. Will the generator find it like this?

whirm:
Yeah, that looks nice, I would mention that it uses VIRTUALENV_DIR too, squash it and I'll merge it.

wouterjsmit:
Done!

PR #251, opened 2016-01-24
Added error if procfs is not available on a system

devos50:
Gummy currently does not support systems where `procfs` is not available.

### DIFF
--- a/run.py
+++ b/run.py
@@ -85,6 +85,10 @@ if __name__ == '__main__':
             print "Error: The specified configuration file doesn't exist."
             exit(1)
 
+        if not exists('/proc'):
+            print "Error: procfs not available on this system."
+            exit(5)
+
         # Create a process group so we can clean up after ourselves when
         setpgrp()  # create new process group and become its leader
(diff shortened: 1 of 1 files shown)


whirm:
It's expected to get IOErrors all the time. (when processes die). I'd rather check for the /proc filesystem to exist at startup than spamming the log with this.

devos50:
@whirm I've added a check for the `/proc` directory instead.

run.py
```diff
@@ -85,6 +85,10 @@ def _killGroup(signal=SIGTERM):
             print "Error: The specified configuration file doesn't exist."
             exit(1)
 
+        if not exists('/proc'):
+            print "Error: procfs not available on this system."
+            exit(1)
```
whirm: Much better, exit with a different value than 1 though, that way we can tell why it failed from the outside if we want to.

devos50: Changed ;)

whirm:
Good, squash and I'll merge.

devos50:
@whirm done

PR #252, opened 2016-01-24
Added conditional import for get_pid_list

### LINKED ISSUE #249: Use psutil.pids() instead of psutil.get_pid_list()
The `psutil.get_pid_list` function has been renamed a long time ago to `psutil.pids()` (see https://pypi.python.org/pypi/psutil vs https://pypi.python.org/pypi/psutil/0.5.0). This makes that we have to install an older version of `psutil`. If we rename this method in the code (there are two occurrences) we can use a more recent version of the `psutil` package.

devos50:
This conditional import adds support for newer versions of `psutil` where the `get_pid_list` method is renamed to `pids`. This fixes #249.

This change should require no changes in other scripts. It has been tested on OS X and Linux with older and newer versions of `psutil`.

I also removed some unused imports.

### DIFF
--- a/run.py
+++ b/run.py
@@ -39,7 +39,7 @@
 
 import logging
 import sys
-from os import environ, getpgid, getpid, getppid, kill, setpgrp
+from os import environ, getpgid, getpid, kill, setpgrp
 from os.path import dirname, exists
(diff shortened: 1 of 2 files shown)


whirm:
Great, thanks!

PR #264, opened 2016-04-13
Continue if any test fails

whirm:


devos50:
Just noticed we aren't print the last 20 lines of the log on OS X/Windows. Should we do it there too?

whirm:
maybe

PR #266, opened 2016-04-23
Collecting network statistics.

jdonkervliet:
Collecting statistics from proc/net/dev

Untested.

### DIFF
--- a/experiments/dummy/local_processguard.conf
+++ b/experiments/dummy/local_processguard.conf
@@ -3,6 +3,6 @@
 
 experiment_name = LocalProcessGuard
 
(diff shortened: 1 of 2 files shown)


devos50:
add to whitelist

whirm:
add to whitelist

jdonkervliet:
I'll change the code to send the network output to a different file.
This makes sense because this data is not gathered per process but globally.
Sending the output to a different file also makes it more difficult to break the current code for the plots.

jdonkervliet:
Done. A file called 'network-usage.log' is now used to monitor network traffic on all available network devices.
Network monitoring is only performed when the network flag (-n, --network) is passed to the process guard.

whirm:
Looks good, can you squash it?

Will you update the plotting part in a future PR?

jdonkervliet:
Squashed.

I would like to have the plots, so I'll update that in a later PR.

whirm:
Great!

PR #275, opened 2016-07-28
WIP: Async dispersy

lfdversluis:
Dispersy will soon return deferreds, this PR is to upgrade parts of Gumby to handle that.

Additionally it features:
- automatically creates a side-by-side comparison graph for send and receive when upstream data is available (through jenkins).


whirm:
add to whitelist

PR #287, opened 2016-12-13
Process historical dropped messages count

kc1212:
I plan to use https://wiki.jenkins-ci.org/display/JENKINS/Copy+Artifact+Plugin to retrieve data from previous successful builds (it's also what Laurens used).

The script works locally, the next step is to test it on Jenkins.

This is my first PR of actual code, so please let me know if you have any comments on the coding style/structure @devos50. Perhaps I should put the dropped message count specific stuff into its own class.

### DIFF
--- a/scripts/post_process_dispersy_experiment.sh
+++ b/scripts/post_process_dispersy_experiment.sh
@@ -100,5 +100,8 @@ export XMIN
 export XMAX
 graph_data.sh
 
+#Step 6: Check historical data for large deviation
+post_process_historical_values.py
+
 #
 # post_process_dispersy_experiment.sh ends here
--- a/scripts/post_process_historical_values.py
+++ b/scripts/post_process_historical_values.py
@@ -1,0 +1,150 @@
+#!/usr/bin/env python2
+
+import csv
+import json
+import os
+import sys
+import numpy as np
+
+
+def is_float(x):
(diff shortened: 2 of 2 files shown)


devos50:
ok to test

kc1212:
@devos50 ready to be reviewed
some parameters need to be tuned, see the `DroppedStatsProcessor.process` method

devos50:
@kc1212 looks ok.

Have you tested it by artificially creating a run with a huge number of dropped messages?

Also, please squash the commit history.

kc1212:
@devos50 squashed the commit history
I tested it on an artificial `dropped_stats.json` that contains a huge number of dropped messages and it works fine. Is that okay or should I look into simulating dropped messages on the test clients (if that's possible)?

kc1212:
I should mention that the Jenkins job needs to be amended for this to work, namely it needs an additional build step that copies the `dropped_stats.json` artifact from the previous build.

devos50:
@kc1212 no need to do that. If you manually verified it, it's ok.

I will merge this PR. Can you change the AllChannels job and add the additional build step? Please don't run it again to save DAS5 nodes (it will run automatically tonight).

kc1212:
@devos50 thanks, I'll do that now

PR #299, opened 2017-02-21
Multichain experiment

devos50:
In this PR, I created a standalone multichain experiment. I reused some code from @Captain-Coder for this.

In addition, I've refactored the tribler client file to create a nice foundation to run Tribler on the DAS5. This is done in the first commit of this PR.

The experiment itself can be found here: https://jenkins.tribler.org/job/Experiment_multichain_integration/. Note that there is no graph created at the end, however, the blocks are written to the `multichain.dat` file so developers can get some insight in the block creation. This experiment will be extended when true halves has been reviewed and merged by me.


experiments/multichain/database_reader.py
```diff
@@ -0,0 +1,130 @@
...
+    def __init__(self, working_directory):
+        """
+        Parent class that contains functionality for both gumby and single database readers.
+        DatabaseReaders process experimentation information.
+        """
+        # Either contains all information all ready or all information will be aggregated here.
+        self.working_directory = working_directory
+        self.database = self.get_database(self.working_directory)
+
+        self.combine_databases()
```
egbertbouman: Maybe you could move these 2 calls outside of the `__init__` and call them after the DatabaseReader is constructed. This way the `__init__` doesn't do any processing.

egbertbouman (changes requested):
Other than this one small remark everything looks good!

PR #301, opened 2017-02-25
Modular community loading

qstokkink:
This resolves #300 and #298. Depends on https://github.com/Tribler/tribler/pull/2815 .

This allows experiment creators to easily isolate and/or overwrite default communities with their own, through community launch profiles (CommunityLauncher objects).

### Hooks
Adds subclass `GumbySession` of `Session` which maintains a subclass `GumbyLaunchMany` of `TriblerLaunchMany`.
The `GumbyLaunchMany` overwrites the `load_communities()` method to use a `CommunityLoader` to `dispersy.define_auto_load` its communities (instead of doing this directly in `load_communities()`).
By loading this into `TriblerExperimentScriptClient`, this allows for experiment writers to easily isolate and/or overwrite default communities.

### New functionality
The default community launch behavior has been included in this PR.
In other words, running `TriblerExperimentScriptClient` will function exactly the same as the hardcoded system.
Due to loading an `IsolatedCommunityLoader` by default this allows for easy isolation of registered communities, by calling `isolate()` for a loaded community, for example:

```python
class MyTriblerExperimentScriptClientSubclass(TriblerExperimentScriptClient):

    def create_community_loader(self):
        loader = super(MyTriblerExperimentScriptClientSubclass, self).create_community_loader()
        loader.isolate("HiddenTunnelCommunity")
        return loader
```

For more drastic changes, one can even replace existing communities by means of creating a custom launcher, for example:

```python
class MyHiddenCommunityLauncher(HiddenTunnelCommunityLauncher):

    def get_community_class(self):
        return MyNewHiddenCommunityClass

class MyTriblerExperimentScriptClientSubclass(TriblerExperimentScriptClient):

    def create_community_loader(self):
        loader = super(MyTriblerExperimentScriptClientSubclass, self).create_community_loader()
        loader.set_launcher(MyHiddenCommunityLauncher)
        return loader
```

Or insert new communities:

```python
class MyNewCommunityLauncher(CommunityLauncher):

    def get_name(self):
        return "MyNewCommunity"

    def get_community_class(self):
        return MyNewCommunityClass

class MyTriblerExperimentScriptClientSubclass(TriblerExperimentScriptClient):

    def create_community_loader(self):
        loader = super(MyTriblerExperimentScriptClientSubclass, self).create_community_loader()
        # Insert a new community
        loader.set_launcher(MyNewCommunityLauncher)
        # Which you can isolate afterwards
        loader.isolate("MyNewCommunity")
        return loader
```

[...]

PR #303, opened 2017-03-06
WIP: Refactored Gumby base classes and experiment setup

Captain-Coder:
The ExperimentClient/DispersyExperimentClient/TriblerExperimentClient inheritance architecture has been revised. The changes are best explained in the following order:

The main() methods of all the different experiment clients have been removed in favor of a single entry point (for an experiment client). This new location is launch.py.

launch.py first checks and sets all the environment variables, this gets rid of various other hacks (BASE_DIR, path(__file__, .., ..., .., 'tribler'), etc)

Most importantly the SCENARIO_FILE environment variable must be set (launch.py will make a few guesses if it is not specified).

Next launch.py sets up logging and instrumentation. Logging was not harmonized for all the experiment clients. Where applicable prints to stdout and stderr and logging.(error|warn|info|debug) calls have been replaced with instanced loggers.

Finally launch sets up an ExperimentClient on the reactor that connects to the sync host.

The ExperimentClient loads the scenario file from the enviroment variable. This scenario file configures the further parts (called modules) that the experiment is composed of. Using a #module instruction in the scenario, one simply names the class that is to be loaded. (Fully qualified with a python module name or an import relative to the scenario file) For common bits of scenario initialization, an #include directive is also supported (see hidden services scenario's for example use)

All modules inherit from the ExperimentModule base class. An inheritor can apply the @static_module decorator. This decorator instances a singleton of this module. Otherwise the classmethod on_module_load must be overriden create actual instances of the module. Each instanced module is passed to the ExperimentClient, that scans the module for method members marked with the @experiment_callback decorator. Each such method is registered as a callback that can be used in the scenario.

Currently there are 2 classes that derive directly from ExperimentModule: BaseDispersyModule and CommunityExperimentModule. The base dispersy module contains all the common elements for the sources of dispersy instances (setting the dispersy port, resetting dispersy statistics, etc.). The 2 inheritors are DispersyModule and TriblerModule, that compare to the old dispersyexperimentclient and triblerexperimentclient. The CommunityExperimentModule is the base class for experiment modules that implement experimentation methods for a specific community class. The role (and functionality) is roughly comparable to the old experiment client base classes (minus the dispersy management stuff). It is the class most implementors will want to use.

This structure allows for good backwards compatibility. The changes to the scenario files themselves are mostly search/replace. The CommunityExperimentModule is not 100% a drop in base replacement. Most changes here are textual too .community as opposed to ._community. But some things that the experiment modules did with hack'n'slash are now possible with more elegant mechanics. So most experiment modules benefitted from a general cleanup and revision. So I included that too.

This PR also includes @qstokkink 's custom community loader. It integrates nicely with the new module system. Each CommunityExperimentModule has access to the community launcher that will be used to lanch the relevant community. The inheritors of CustomExperimentModule can use this to control very precisely how/when their community is loaded. Another usefull feature of qstokkink's work is community isolation. Instead of setting a master member for each community in the various scenario's, it is now possible to isolate_community in the scenario file. It produces a master member for the isolated community that is stable for all the nodes in the experiment, but is very unlikely to clash across multiple instances of the experiment being run concurrently.

It also turned out that Mulitchain was not the first experiment to use multiple communities. The BarterCommunity experiment also requires multiple communities to be loaded and controlled at the same time. After the module system refactor this turned out nicely separated. The bartercast scenario just imports all the needed CommunityExperimentModules and calls what it needs.

There are several experiment clients that experiment with communities that are not (no longer or never where) in tribler. I removed these as there is little point in keeping them around. If somone needs them, they are in the git history.

Further minor changes:
- I've tried to PEP8 the obvious/bad failures, but there was far to much crap to fix it all
- rename everything to underscore_identifiers where possible (twisted has camelCase it seems, so these could not be renamed). 
- Unified the dispersy port used, such that the experimentserver now actually produces all_var structures that contain the right host/port combo's in all cases.
- I tweaked the experimentclient protocol slightly, such that it is now (reliably) possible to set the experimentclient's var structure based on the id assigned.

There is some detail work, testing and commit restructuring to be done, so WIP for now.


qstokkink:
My code portion has been reviewed and **rebased** already in #301. Also wow, 1 million LOC changed.

devos50:
ok to test

devos50:
@Captain-Coder thanks! I will review it tomorrow.

experiments/__init__.py
```diff
@@ -0,0 +1 @@
+
```
devos50: Please add a small description here what can be found in the package.

experiments/dispersy/__init__.py
```diff
@@ -0,0 +1 @@
+
```
devos50: Please add a small description here what can be found in the package.

gumby/launch.py
```diff
@@ -0,0 +1,77 @@
...
+from os import environ, path
+from random import random
+from sys import path as python_path, argv
+
+from gumby.instrumentation import init_instrumentation
+from gumby.log import setupLogging
+from gumby.sync import ExperimentClientFactory, ExperimentServiceFactory
+
+
+def main(self_service=False):
```
devos50: Please add some general comments here what this file does (like you explained in the first post).

gumby/launch.py
```diff
@@ -0,0 +1,77 @@
+from logging import debug, error
```
devos50: I think the name of this file is a bit too vague. Something like `launch_experiment.py` is better IMO.

gumby/launch.py
```diff
@@ -0,0 +1,77 @@
...
+from gumby.sync import ExperimentClientFactory, ExperimentServiceFactory
+
+
+def main(self_service=False):
+    if len(argv) > 2:
+        print "Launch invoke error, to many command line arguments. Specify 1 scenario file to run."
+        return
+
+    if len(argv) == 2:
+        p = path.abspath(argv[1])
```
devos50: Please don't use one-letter variable names.

gumby/launch.py
```diff
@@ -0,0 +1,77 @@
...
+        if path.exists(p):
+            print "Launch error, command line argument %s does not exist" % p
+        elif "SCENARIO_FILE" in environ:
+            print "Launch invoke error, can't take both a command line scenario file and ean nvironment scenario file."
```
devos50: Typo at the end of this sentence.

gumby/launch.py
```diff
@@ -0,0 +1,77 @@
...
+    if "SCENARIO_FILE" not in environ:
+        environ["SCENARIO_FILE"] = path.abspath(path.join(
+            environ["EXPERIMENT_DIR"], "%s.scenario" % path.basename(path.normpath(environ["EXPERIMENT_DIR"]))))
+
+    python_path.append(environ["TRIBLER_DIR"])
+
+    init_instrumentation()
+    setupLogging()
+    if not path.exists(environ["SCENARIO_FILE"]):
+        error("Unable to find scenario file: %s", environ["SCENARIO_FILE"])
```
devos50: I'm not sure whether we should require a scenario file before running an experiment. I rather have the scenario file as optional feature since I can imagine experiments that do not necessarily require scenarios (like the Tribler idle runner). @qstokkink and/or @egbertbouman what's your opinion on this?

EDIT: apparently, the `ExperimentClient` class can handle the case when there's no scenario file.

gumby/launch.py
```diff
@@ -0,0 +1,77 @@
...
+
+    reactor.exitCode = 0
+
+    # if self service is requested, start an experiment server to run our scenario. Used to debug the client in an IDE
+    if self_service or "GUMBY_SELF_SERVICE" in environ:
+        environ["SYNC_HOST"] = "localhost"
+        if environ["SYNC_PORT"] == "0":
+            environ["SYNC_PORT"] = "57756"       # corresponds to the keyboard rows for the word gumby
```
devos50: What kind of keyboard do you have lol?

gumby/launch.py
```diff
@@ -0,0 +1,77 @@
...
+        debug("Starting experiment server on: %s:%s", environ['SYNC_HOST'], int(environ['SYNC_PORT']))
+
+        def exp_started(_):
+            debug("Experiment started, closing server")
+
+        fact = ExperimentServiceFactory(1, 3)
```
devos50: What are these 1, 3 parameters?

gumby/launch.py
```diff
@@ -0,0 +1,77 @@
...
+        reactor.listenTCP(int(environ['SYNC_PORT']), fact)
+
+    debug("Connecting to: %s:%s", environ['SYNC_HOST'], int(environ['SYNC_PORT']))
+    reactor.callLater(1 + random() * 10, reactor.connectTCP,
+                      environ['SYNC_HOST'], int(environ['SYNC_PORT']), ExperimentClientFactory({}))
+    reactor.run()
+    exit(reactor.exitCode)
+
+if __name__ == '__main__':
+    main()
```
devos50: Missing newline at the end of the file.

gumby/sync.py
```diff
@@ -337,39 +365,68 @@ def __init__(self, vars):
         self.vars = vars
         self.all_vars = {}
         self.time_offset = None
+        self.scenario_runner = ScenarioRunner()
+        self.scenario_runner.preprocessor_callbacks["module"] = self._preproc_module
+        self.loaded_experiment_module_classes = []
+        self.experiment_modules = []
+        self._stats_file = None
+        if not hasattr(self, 'scenario_file'):
```
devos50: Why is this `if not` necessary?

gumby/sync.py
```diff
@@ -337,39 +365,68 @@ def __init__(self, vars):
...
+            self.scenario_file = environ.get("SCENARIO_FILE", None)
+
+        self.register_callbacks(self)
+
+        if self.scenario_file is None:
+            self._logger.error("No Scenario file defined, starting empty experiment")
```
devos50: Scenario doesn't have to be capitalised.

gumby/sync.py
```diff
@@ -304,26 +333,25 @@ def unregisterConnection(self, proto):
...
 
     def lineLengthExceeded(self, line):
         self._logger.error("Line length exceeded, %d bytes remain.", len(line))
+
 #
 # Client side
 #
-
-
 class ExperimentClient(object, LineReceiver):
```
devos50: Can you move this class to a separate file?

gumby/sync.py
```diff
@@ -337,39 +365,68 @@ def __init__(self, vars):
...
+        if path.exists(my_dir):
+            self._logger.warning("Output directory already exists, should you clean before experiment? (%s)", my_dir)
+        else:
+            makedirs(my_dir)
+        chdir(my_dir)
+        self._stats_file = open("statistics.log", 'w')
 
-    def onAllVarsReceived(self):
-        self._logger.debug("onAllVarsReceived: Call not implemented")
+        for m in self.experiment_modules:
```
devos50: No one-letter variable names please.

experiments/dispersy/allchannel_1000.scenario
```diff
@@ -1,6 +1,8 @@
-@0:0 set_master_member 3081a7301006072a8648ce3d020106052b81040027038192000403cbbfd2dfb67a7db66c88988df56f93fa6e7f982f9a6a0fa8898492c8b8cae23e10b159ace60b7047012082a5aa4c6e221d7e58107bb550436d57e046c11ab4f51f0ab18fa8f58d0346cc12d1cc2b61fc86fe5ed192309152e11e3f02489e30c7c971dd989e1ce5030ea0fb77d5220a92cceb567cbc94bc39ba246a42e215b55e9315b543ddeff0209e916f77c0d747
-@0:0 start_dispersy
-@0:1 online {1-980,990-1000}
+#module gumby.modules.dispersy_module.DispersyModule
```
devos50: While the idea of preprocessor directives here is nice, I think it's too confusing since we also support comments. Can you think of another symbol here?

gumby/modules/experiment_module.py
```diff
@@ -0,0 +1,56 @@
+import logging
+
+
+class ExperimentModule(object):
```
devos50, line 4: Please add a small explanation of this class in comments.

gumby/modules/experiment_module.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def on_id_received(self):
+        pass
+
+    @staticmethod
+    def str2bool(v):
+        return v.lower() in ("yes", "true", "t", "1")
+
+    @staticmethod
+    def str2tuple(v):
```
devos50: Is this still used?

gumby/modules/experiment_module.py
```diff
@@ -0,0 +1,56 @@
...
+        if len(v) > 1 and v[1] == "t":
+            return (int(v[0]), int(v[2:]))
+        if len(v) > 1 and v[1] == ".":
+            return float(v)
+        return int(v)
+
+
+def static_module(cls):
```
devos50: Please add some comments here on why this decorator is necessary and what it does exactly.

gumby/scenario.py
```diff
@@ -87,59 +85,75 @@ class ScenarioParser():
...
+        self.line_buffer = []
+        self.preprocessor_callbacks = {
+            "include": self._preproc_include_file
+        }
 
-    def _read_scenario(self, filename):
+    def add_scenario(self, filename):
         """
-        Read the scenario into memory return a list containing the lines.
+        Read the scenario into this scenario parser
```
devos50: Please end the sentence with a dot.

gumby/scenario.py
```diff
@@ -87,59 +85,75 @@ class ScenarioParser():
                time stamp, they will be executed in order.
     """
     _re_substitution = re_compile("(\$\w+)")
+    _re_preprocessor_dir = re_compile("^#(\w+)\s+")
+    _re_named_arg = re_compile("^\s*(\w+)\s*=\s*(.*)$")
```
devos50: I understand the added benefit of named arguments but I would also argue that they are not really necessary for now and could be added in a subsequent PR 👍 .

gumby/scenario.py
```diff
@@ -87,59 +85,75 @@ class ScenarioParser():
...
-                    if not line.startswith('#'):
-                        line = line.strip()
-                        line_buffer.append((linenr, line))
-                    linenr += 1
-
-                self.file_buffer = (filename, line_buffer)
-        return self.file_buffer[1]
-
-    def _parse_scenario(self, filename):
+            f = open(filename, "r")
```
devos50: Please use `with open(filename, "r") as scenario_file:`.

gumby/scenario.py
```diff
@@ -299,25 +312,4 @@ def _parse_for_this_peer(self, peerspec):
...
-
-    t1 = time()
-    sr = ScenarioRunner(sys.argv[1])
-    sr.set_peernumber(peer_id)
-    sr.parse_file()
-
-    print >> sys.stderr, "Took %.2f to parse %s" % (time() - t1, sys.argv[1])
-    for tstmp, clb, args in sr._my_actions:
-        print >> sys.stderr, tstmp, clb, args
+# scenario.py ends here
```
devos50: Please add a missing newline.

gumby/scenario.py
```diff
@@ -87,59 +85,75 @@ class ScenarioParser():
...
-    def _parse_scenario(self, filename):
+            f = open(filename, "r")
+            lines = f.readlines()
+            f.close()
+
+            line_number = 1
+            for line in lines:
+                if not line.startswith('#'):
+                    line = line.strip()
+                    self.line_buffer.append((filename, line_number, line))
```
devos50: You're passing the same filename over and over again to the `_parse_scenario` method. Can this be avoided?

gumby/scenario.py
```diff
@@ -209,21 +233,21 @@ def _parse_peerspec(self, peerspec):
...
                 line = line.replace(substitution, environ[substitution[1:]])
 
         return line
 
-class ScenarioRunner(ScenarioParser):
 
+class ScenarioRunner(ScenarioParser):
     """
-    Reads, parses and schedules events from scenario file.
+    Reads, parses and schedules events from scenario files.
```
devos50, line 252: Does this mean that we are supporting multiple scenario files in one experiment?

gumby/sync.py
```diff
@@ -61,18 +61,45 @@
...
 
+
+#
+# Aux stuff
+#
+
+
+def stop_reactor():
+    if reactor.running:
+        logger.debug("Stopping reactor")
```
devos50: I think this should be on info level.

gumby/sync.py
```diff
@@ -61,18 +61,45 @@
...
+        reactor.stop()
+
+
+def experiment_callback(name=None):
```
devos50: Please add comments here to explain what this decorator is doing.

gumby/sync.py
```diff
@@ -235,8 +262,10 @@ def pushInfoToSubscribers(self):
         vars = {}
         for subscriber in self.connections_ready:
             subscriber_vars = subscriber.vars.copy()
-            subscriber_vars['port'] = subscriber.id + 12000
-            subscriber_vars['host'] = subscriber.transport.getPeer().host
+            if "port" not in subscriber_vars:
```
devos50, line 233: Why is this if-statement necessary?

gumby/sync.py
```diff
@@ -419,10 +476,99 @@ def proto_go(self, line):
...
+
+    @experiment_callback
+    def peertype(self, peer_type):
+        self._stats_file.write('%.1f %s %s %s\n' % (time(), self.my_id, "peertype", peer_type))
+
+    @experiment_callback
+    def stop(self):
+        stop_reactor()
+
+    def print_on_change(self, name, prev_dict, cur_dict):
```
devos50: I think this method is very vague, both the name and the functionality. If I understood it correctly, it writes a dictionary as soon as a value in it changes. I would rather name this method `print_dict_on_change` or something.

gumby/sync.py
```diff
@@ -419,10 +476,99 @@ def proto_go(self, line):
...
+            return new_values, changed_values
+
+        new_values, changed_values = get_changed_values(prev_dict, cur_dict)
+        if changed_values:
+            self._stats_file.write('%.1f %s %s %s\n' % (time(), self.my_id, name, json.dumps(changed_values)))
+            self._stats_file.flush()
+            return new_values
+        return prev_dict
+
+    def _preproc_module(self, filename, line_number, line):
```
devos50: Please add some comments here, this method is quite complex and requires some explanation.

experiments/dispersy/allchannel_1000.scenario
```diff
@@ -1,6 +1,8 @@
...
+#module experiments.dispersy.allchannel_module.AllChannelModule
+
+@0.0 isolate_community AllChannelCommunity
```
devos50: Isn't this loading 1000 different isolated communities?

gumby/modules/dispersy_module.py
```diff
@@ -0,0 +1,166 @@
...
+        prev_total_received = {}
+        prev_total_dropped = {}
+        prev_total_delayed = {}
+        prev_total_outgoing = {}
+        prev_total_fail = {}
+        prev_endpoint_recv = {}
+        prev_endpoint_send = {}
+        prev_created_messages = {}
+
+        while True:
```
devos50: Please use a `LoopingCall` instead.

gumby/modules/base_dispersy_module.py
```diff
@@ -0,0 +1,70 @@
...
+    @experiment_callback
+    def reset_dispersy_statistics(self):
+        self.dispersy._statistics.reset()
+
+    @experiment_callback
+    def isolate_community(self, name):
+        self.custom_community_loader.isolate(name)
+
+    @experiment_callback
+    def observe_exceptions(self):
```
devos50: This appears to be unused?

gumby/modules/community_experiment_module.py
```diff
@@ -0,0 +1,58 @@
+from gumby.modules.experiment_module import ExperimentModule
+
+
+class CommunityExperimentModule(ExperimentModule):
```
devos50: Please add some comments on how to use this module.

gumby/modules/community_experiment_module.py
```diff
@@ -0,0 +1,58 @@
...
+        return self.dispersy_provider.custom_community_loader
+
+    @property
+    def community_launcher(self):
+        return  self.community_loader.get_launcher(self.community_class.__name__)
+
+    @property
+    def community(self):
+        for c in self.dispersy.get_communities():
+            if isinstance(c, self.community_class):
```
devos50: What happens if there are multiple communities with the same type loaded (i.e. very common for `ChannelCommunity`)?

gumby/modules/community_experiment_module.py
```diff
@@ -0,0 +1,58 @@
...
+    def community_loader(self):
+        return self.dispersy_provider.custom_community_loader
+
+    @property
+    def community_launcher(self):
+        return  self.community_loader.get_launcher(self.community_class.__name__)
+
+    @property
+    def community(self):
+        for c in self.dispersy.get_communities():
```
devos50: Please don't use one-letter variable names.

gumby/modules/community_experiment_module.py
```diff
@@ -0,0 +1,58 @@
...
+        self.on_community_loaded()
+
+    def on_community_loaded(self):
+        pass
+
+    def set_community_launcher(self, community_launcher):
+        loader = self.community_loader
+        if loader is None:
+            self._logger.error("Unable to set community launcher, no custom community loader is active")
+        self.community_loader.set_launcher(community_launcher)
```
devos50: Please add a newline.

gumby/modules/dispersy_module.py
```diff
@@ -0,0 +1,166 @@
...
+    @experiment_callback
+    def stop_session(self):
+        self.dispersy.stop()
+
+    @experiment_callback
+    def churn(self, *args):
+        self.print_on_change('community-churn', {}, {'args':args})
+
+    def _initialize_crypto(self):
+        if environ.get('TRACKER_CRYPTO', 'ECCrypto') == 'ECCrypto':
```
devos50: We should change the name `TRACKER_CRYPTO` to something like `DISPERSY_CRYPTO` since it does not only affect the crypto used by the trackers.

gumby/modules/tribler_module.py
```diff
@@ -0,0 +1,42 @@
...
+from gumby.sync import experiment_callback
+from gumby.modules.experiment_module import static_module
+from gumby.modules.base_dispersy_module import BaseDispersyModule
+
+
+@static_module
+class TriblerModule(BaseDispersyModule):
+    def __init__(self, experiment):
+        super(TriblerModule, self).__init__(experiment)
+        # in the case of Tribler dispersy we cannot control the launch of the discovery community. So don't try it.
```
devos50: This sentence reads a bit weird.

gumby/modules/tribler_module.py
```diff
@@ -0,0 +1,42 @@
...
+            self.dispersy = self.session.lm.dispersy
+
+        return self.session.start().addCallback(on_tribler_started)
+
+    @experiment_callback
+    def stop_session(self):
+        deferToThread(self.session.shutdown)
+
+    def setup_session_config(self):
+        # symlink the bootstrap file so we are only connecting to our own trackers
```
devos50: We don't always want to connect to only our trackers (i.e. when we run Tribler idle) so we should make this optional.

experiments/dispersy/allchannel_module.py
```diff
@@ -36,72 +36,55 @@
...
-        self._dispersy.define_auto_load(ChannelCommunity, self._my_member, (), {"tribler_session": None})
-        self._dispersy.define_auto_load(PreviewChannelCommunity, self._my_member, (), {"tribler_session": None})
-
-    def create(self):
+    @experiment_callback
+    def allchannel_create(self):
         self._logger.info("creating-community")
-        self.my_channel = ChannelCommunity.create_community(self._dispersy, self._my_member, tribler_session=None)
+        self.my_channel = ChannelCommunity.create_community(self.dispersy, self.community.my_member,
+                                                            tribler_session=self.session)
```
devos50: Regarding the AllChannel experiment, we should make sure that we are still using an in-memory database now that we are passing a Tribler session to the `ChannelCommunity` (same holds for `AllChannel`). Having 1000 clients writing everything to db files, will only slow down things.

experiments/dispersy/hidden_services_module.py
```diff
@@ -0,0 +1,207 @@
...
+
+    @experiment_callback
+    def introduce_candidates(self):
+        for port in range(12000, 12000 + self.total_peers):
+            if self.dispersy_provider.dispersy_port != port:
+                self.community.add_discovered_candidate(Candidate((self.dispersy.wan_address[0], port),
+                                                                   tunnel=False))
+
+    @experiment_callback
+    def start_download(self, filename, hops=1):
```
devos50: I think this method should be located in `TriblerModule`.

devos50 (changes requested):
First of all, thank you for your hard work! I think Gumby becomes a lot more usable after this.

I’ve reviewed the PR and I have some general comments/questions. In general, I’m satisfied with the changes but it took me a while to get familiar with all the new concepts. When we discussed the design, it was indicated that a minimal-changes approach is the most suitable for now. While I understand the cascading effect when refactoring, I think some of these changes were not really necessary for now. This would also allow you to create smaller PRs and make more incremental changes. Please keep this in mind the next time you refactor.

Another general comment is that there are many design decisions made without discussing this with any of us (modules, preprocessor directives). While these features without doubt make Gumby better, I think it would be good that we discussed the major new functionality prior to implementing it so we have a better understanding of where we are going as a team.

I like the modular loading idea and this is a very flexible way of loading multiple communities at once. I’m curious to know how this turns out.

Testing:
Since this might affect many of our experiments, I would like to verify that the code is working correctly by running the AllChannel, Multichain, and Tribler idle run experiments on this branch before merging this. You can simply copy the Jenkins jobs and change the Gumby repo to test it.

Documentation:
Our new way of running experiments and creating scenario files requires some explanation. Quinten already created a .rst file in his PR. Do you mind to write some basic documentation on how to run experiments using Tribler and/or Dispersy in .rst format?

Pylint:
I’m thinking of setting up a Pylint job for Gumby, however, for now I don’t really mind about existing Pylint errors. Just try to fix most of the obvious errors (like missing newlines at the end of files) if you see them.

Also, I will merge #301 before this. After that, you can rebase your PR.

@egbertbouman and @qstokkink if you have time, please also give your comments on this PR :)

gumby/launch.py
```diff
@@ -0,0 +1,77 @@
...
+from sys import path as python_path, argv
+
+from gumby.instrumentation import init_instrumentation
+from gumby.log import setupLogging
+from gumby.sync import ExperimentClientFactory, ExperimentServiceFactory
+
+
+def main(self_service=False):
+    if len(argv) > 2:
+        print "Launch invoke error, to many command line arguments. Specify 1 scenario file to run."
```
qstokkink: "too many"

gumby/launch.py
```diff
@@ -0,0 +1,77 @@
...
+        return
+
+    if len(argv) == 2:
+        p = path.abspath(argv[1])
+        if path.exists(p):
+            print "Launch error, command line argument %s does not exist" % p
```
qstokkink: Shouldn't this return?

gumby/launch.py
qstokkink: Even the Tribler idle needs to know how long it is supposed to run. Force-killing it through the process_guard/DAS5 timeout is not really ideal design. I'd vote for a required definition of some scenario file.

experiments/dispersy/hidden_services_module.py
qstokkink: I agree, but should that refactoring be part of this PR?

experiments/dispersy/allchannel_1000.scenario
qstokkink: Maybe `$` or change the comment symbol to `%`.

gumby/modules/dispersy_module.py
```diff
@@ -0,0 +1,166 @@
...
+from Tribler.dispersy.candidate import CANDIDATE_STUMBLE_LIFETIME, CANDIDATE_WALK_LIFETIME, CANDIDATE_INTRO_LIFETIME
+
+
+@static_module
+class DispersyModule(BaseDispersyModule):
+    def __init__(self, experiment):
+        super(DispersyModule, self).__init__(experiment)
+        self._crypto = self._initialize_crypto()
+        self.session_id = environ['SYNC_HOST'] + environ['SYNC_PORT']
+        self.custom_community_loader = self.create_community_loader()
```
qstokkink: Wasn't this already defined in the `BaseDispersyModule`? [Comment applies for all other BaseDispersyModule subclasses as well]

qstokkink (changes requested):
I agree with most of @devos50's comments (for a select few I have added my thoughts).

Overall, this could very much use (at least) class level comments to tell the reviewer/user the responsibilities and the capabilities of the class. 

Other than that, good job 👍.

gumby/launch.py
Captain-Coder: Alternativey, we could rename launch.py to launch_scenario.py. That makes it clear that this launcher is to be used for scenario's only. For experiments that have no need of scenario's, they can still derive from ExperimentClient and do what ever they need to. But they will then need a custom launcher.

gumby/sync.py
Captain-Coder: If "port" is in subscriber vars, that means that the experiment client/node is explicitly telling the experiment server it is (going to be) using that port. With the old behaviour the experiment server is just ignoring what the client said and communicating it's thoughts on the matter to the other nodes. I'm unsure if it ever happened, but the old tribler experiment client used ports 21000 and up, so using a var[client_id]["port"] to construct a candidate should fail. At any rate, having the experiment server mess with my variables would not be something I'd expect as user of the vars-system.

gumby/sync.py
Captain-Coder: It's a relic. Inheritors of ExperimentClient would set this variable to indicate the scenario file they'd want to run. I kept it there at first and apprently forgot about it.

gumby/sync.py
Captain-Coder: This is a relic. It came from the old dispersy experiment client. Some of the older scenarios use it. They keep a dict, and periodically (usually a LoopingCall), construct a new dict and pass the old and new in. Then some changed variables would get printed and the new one stored as old for the next cycle. I dunno how usefull or not it is.

As to the naming, you're right, we might as well change it. The `_on_` term suggests an event or something automatic, which this is not. So perhaps `print_dict_changes`?

experiments/dispersy/allchannel_1000.scenario
Captain-Coder: The thought had occured to me too. The `$` is no good, since we do environment variable expansion based on that. I used the `#` because I expect more people to associate it with preprocessor directives. On the one hand its easy since we only have 1 symbol that does something magic, the less we have of those the better imo. But yes it does impact/confuse comments. Also I found I don't like the # and m next to oneanother. Too many vertical-ish stripes. ~module ?module &module %module all look good to me.

gumby/scenario.py
Captain-Coder: I wanted to keep the filename with the line number. That way errors ("unregistered callback used", "module import error" or "include file doesn't exist") can tell you exactly where the problem is.

gumby/scenario.py
Captain-Coder: Well you can #include them, so even though they form one whole, they come from multiple file.
I thought about having the "SCENARIO_FILE" environment variable as a list of files, but I was undecided on that one.

experiments/dispersy/allchannel_1000.scenario
qstokkink: Out of those I'd pick `?` or `&`:

`?` reads as `unknown`, sparking the intuition that it is unkown to this file and needs to be found elsewhere
`&` reads as `and`, intuitively this seems like an additional element to be included in this file

gumby/modules/experiment_module.py
Captain-Coder: Yes, it is used to parse True/False style arguments to the scenario callback functions. For example `should_become_exit` in the hidden services module uses it.

experiments/dispersy/allchannel_1000.scenario
Captain-Coder: As far as I can tell, no. The AllChannelCommunity itself is isolated, however the communities that it spawns are loaded directly into dispersy.

gumby/modules/dispersy_module.py
Captain-Coder: After grepping, this turned out the be the only use for TRACKER_CRYPTO, appart from the obvious setting of the crypto of the tracker. I think this is mostly to make the dispersy trackers compatible with the dispersy instances in the experiment clients. So considering this, it's not the worst problem in the world. I will change this to an experiment callback that can be used to configure what crypto to use when starting dispersy.

gumby/modules/community_experiment_module.py
Captain-Coder: Yup, that's a pickle. Well sort of. It requires a bit of code, but I thought it best to leave that as future work, since it is not directly required for this PR. I added a todo comment here with the outline of how to handle that situation. The ChannelCommunity is only in gumby through the AllChannelModule. If future users want to create a ChannelCommunityModule, they run into this problem pretty quickly and read/pickup on this todo comment.

devos50:
@Captain-Coder now that #301 has been merged, please rebase this PR.

qstokkink:
retest this please

qstokkink:
retest this please

qstokkink:
retest this please

qstokkink:
retest this please

qstokkink:
retest this please

qstokkink:
@Captain-Coder https://jenkins.tribler.org/job/pers/job/allchannel_base_refactor/

This is running `./gumby/run.py gumby/experiments/dispersy/allchannel.conf`.

```
WARNING 1493635847.59 launch_scenario:71    Unable to find scenario file: /var/scratch/pouwelse/jenkins/workspace/pers/allchannel_base_refactor/gumby/experiments/dispersy/dispersy.scenario
ERROR   1493635848.68      experiment:79    Scenario file /var/scratch/pouwelse/jenkins/workspace/pers/allchannel_base_refactor/gumby/experiments/dispersy/dispersy.scenario not found
```

By adding `scenario_file = 'allchannel_1000.scenario'` to the conf file I got:

```python
Traceback (most recent call last):
  File "./gumby/run.py", line 103, in <module>
    exp_runner.run()
  File "/var/scratch/pouwelse/jenkins/workspace/pers/allchannel_base_refactor/gumby/gumby/runner.py", line 330, in run
    rmtree(path.join(self._output_dir, element))
  File "/usr/lib64/python2.7/shutil.py", line 239, in rmtree
    onerror(os.listdir, path, sys.exc_info())
  File "/usr/lib64/python2.7/shutil.py", line 237, in rmtree
    names = os.listdir(path)
OSError: [Errno 20] Not a directory: '/var/scratch/pouwelse/jenkins/workspace/pers/allchannel_base_refactor/output/bootstraptribler.txt'
```

Captain-Coder:
@qstokkink Sorry about that. I have a gumby invoke script that sets local stuff such as the propper pythonpath. And cleans the output directory. <- whoops! So I've updated my script to be a little more minimal, and fixed the error. And verified that I can run gumby without my invoke script without any errors.

qstokkink:
@Captain-Coder don't forget to add the scenario file to the allchannel conf.

Output:
```python

Starting #00000: launch_scenario.py
Found a logger.conf, using it.
WARNING 1493716965.24 launch_scenario:71    Unable to find scenario file: allchannel_1000.scenario
WARNING 1493716966.31      experiment:73    Scenario file allchannel_1000.scenario not found, attempting scenario file in experiment dir
ERROR   1493716966.37      experiment:295   Unable to import gumby.modules.dispersy_module.DispersyModule (from /var/scratch/pouwelse/jenkins/workspace/pers/allchannel_base_refactor/gumby/experiments/dispersy/allchannel_1000.scenario:1)
Traceback (most recent call last):
  File "/var/scratch/pouwelse/jenkins/workspace/pers/allchannel_base_refactor/gumby/gumby/experiment.py", line 293, in _preproc_module
    stuff = __import__(line)
  File "/var/scratch/pouwelse/jenkins/workspace/pers/allchannel_base_refactor/gumby/gumby/modules/dispersy_module.py", line 8, in <module>
    from gumby.modules.community_launcher import DiscoveryCommunityLauncher
  File "/var/scratch/pouwelse/jenkins/workspace/pers/allchannel_base_refactor/gumby/gumby/modules/community_launcher.py", line 3, in <module>
    from Tribler.community.search.community import SearchCommunity
ImportError: No module named Tribler.community.search.community
ERROR   1493716966.46      experiment:295   Unable to import experiments.dispersy.allchannel_module.AllChannelModule (from /var/scratch/pouwelse/jenkins/workspace/pers/allchannel_base_refactor/gumby/experiments/dispersy/allchannel_1000.scenario:2)
Traceback (most recent call last):
  File "/var/scratch/pouwelse/jenkins/workspace/pers/allchannel_base_refactor/gumby/gumby/experiment.py", line 293, in _preproc_module
    stuff = __import__(line)
  File "/var/scratch/pouwelse/jenkins/workspace/pers/allchannel_base_refactor/gumby/experiments/dispersy/allchannel_module.py", line 47, in <module>
    from gumby.modules.community_experiment_module import CommunityExperimentModule
  File "/var/scratch/pouwelse/jenkins/workspace/pers/allchannel_base_refactor/gumby/gumby/modules/community_experiment_module.py", line 4, in <module>
    from gumby.modules.base_dispersy_module import BaseDispersyModule
  File "/var/scratch/pouwelse/jenkins/workspace/pers/allchannel_base_refactor/gumby/gumby/modules/base_dispersy_module.py", line 7, in <module>
    from gumby.modules.gumby_session import GumbySession
  File "/var/scratch/pouwelse/jenkins/workspace/pers/allchannel_base_refactor/gumby/gumby/modules/gumby_session.py", line 7, in <module>
    from gumby.modules.community_launcher import *
  File "/var/scratch/pouwelse/jenkins/workspace/pers/allchannel_base_refactor/gumby/gumby/modules/community_launcher.py", line 3, in <module>
    from Tribler.community.search.community import SearchCommunity
ImportError: No module named Tribler.community.search.community
ERROR   1493716969.52        scenario:201   Error reading scenario /var/scratch/pouwelse/jenkins/workspace/pers/allchannel_base_refactor/gumby/experiments/dispersy/allchannel_1000.scenario:4, invalid line @0.0 isolate_community AllChannelCommunity.
Traceback (most recent call last):
  File "/var/scratch/pouwelse/jenkins/workspace/pers/allchannel_base_refactor/gumby/gumby/scenario.py", line 181, in _parse_scenario_line
    begin = int(timespec[-1])
ValueError: invalid literal for int() with base 10: '0.0'
ERROR   1493716969.55        scenario:295   Error running scenario /var/scratch/pouwelse/jenkins/workspace/pers/allchannel_base_refactor/gumby/experiments/dispersy/allchannel_1000.scenario:5, undefined callback start_session.
ERROR   1493716969.55        scenario:295   Error running scenario /var/scratch/pouwelse/jenkins/workspace/pers/allchannel_base_refactor/gumby/experiments/dispersy/allchannel_1000.scenario:6, undefined callback reset_dispersy_statistics.
ERROR   1493716969.55        scenario:295   Error running scenario /var/scratch/pouwelse/jenkins/workspace/pers/allchannel_base_refactor/gumby/experiments/dispersy/allchannel_1000.scenario:66, undefined callback join.
ERROR   1493716969.57        scenario:295   Error running scenario /var/scratch/pouwelse/jenkins/workspace/pers/allchannel_base_refactor/gumby/experiments/dispersy/allchannel_1000.scenario:2292, undefined callback stop_session.
```

This looks like the launchers are importing the Tribler classes, without the Tribler folder being added to the path.

gumby/modules/community_launcher.py
```diff
@@ -1,5 +1,14 @@
 from abc import ABCMeta, abstractmethod
 
+from Tribler.community.search.community import SearchCommunity
```
qstokkink, line 3: This change should be reverted. This may look nicer, but moving the Tribler imports to the module level instead of the class level has two major downsides:

 1. Anyone who happens to be so unfortunate to have this module anywhere in their import chain, needs to have Tribler on their path. This may not seem like a big deal, until someone imports `custom_string from a_random_gumby_module` and has to figure out why he needs Tribler on the path.

 2. When running an experiment for _community X_ you should load the code from the _community X_ module. Now we may explicitly tell the launcher to only load _community X_, but we get the code from _community A_ through _community Z_ as well. Should these imports happen to conflict, the user won't know why: this is exactly what isolation is supposed to prevent.

qstokkink:
These are all the fixes I needed to get this in working order (this includes fixes for #310):
https://github.com/qstokkink/gumby/commit/76181be2ca80cee9e854170598f7853a518fa3c8

Please cherry-pick and squash your history so we can merge this PR. 👍 

P.S. Output for AllChannel experiment: https://jenkins.tribler.org/job/pers/job/allchannel_base_refactor/15/

gumby/modules/community_launcher.py
Captain-Coder: I considered some of this too when I moved them. If your scenario imports any of the dispersy providers (tribler_module or dispersy_module), you will also import GumbySession and thus Session from Tribler. So the chances of needing tribler on your path are 99.9% already. (All scenario experiments to date have this requirement)

A case could be made for moving the GumbySession to only be used by the tribler_module. (it used to be that way in gumby) Some communities behave differently depending on the existance of the session. I thought this undesirable since that means that tribler_module and dispersy_module are not interchangabe. But I agree that it might not be 100% obvious that importing dispersy_module in a scenario also causes an import of Tribler. (But after moving the GumbySession to tribler_module you will also need to move all the session config objects there, since that requires Tribler on your path too. And then the community modules need separate code paths to configure stuff depending on which dispersy provider is loaded)

Regarding point 2. I hadn't considered this at all. It might also save some resources (less parsing, less memory needed) on each instance.

PR #306, opened 2017-03-17
Use the proper keypair for the Multichain Community

pimveldhuisen:
This PR makes the multichain_client.py use the correct keypair. The default TriblerExperimentClient uses a different type of keypair, which prevents Multichain records from being countersigned in the experiment.


devos50 (approved):

devos50:
Thanks!

PR #307, opened 2017-03-23
Remove Bartercast community-related code from non-legacy folders

vandenheuvel:
As the Bartercast community is now [removed](https://github.com/Tribler/tribler/pull/2858) from Tribler it should be removed here as well.


devos50:
ok to test

devos50 (approved):
Thanks!

PR #313, opened 2017-05-10
Fixed dynamic module importing

qstokkink:
This imports modules dynamically with correct paths. This now allows modules to perform imports in the structure. Also adds unit tests for this.


qstokkink:
retest this please

PR #320, opened 2017-05-12
Made extract_dispersy_statistics localhost compatible

### LINKED ISSUE #318: extract_dispersy_statistics only works for DAS5 deployment
[extract_dispersy_statistics.py](https://github.com/Tribler/gumby/blob/devel/experiments/dispersy/extract_dispersy_statistics.py) expects the folders to be layed out like in the DAS5 deployment.

This should be made compatible with localhost experiments:
https://github.com/Tribler/gumby/blob/devel/experiments/dispersy/extract_dispersy_statistics.py#L151

qstokkink:
Fixes #318.

- Instead of only yielding files which conform to the DAS output folder structure, extract_dispersy_statistics.py also yields/finds localhost output files.
- Also checks if the log files are non-empty, which would cause trouble further into the script.

### DIFF
--- a/experiments/dispersy/extract_dispersy_statistics.py
+++ b/experiments/dispersy/extract_dispersy_statistics.py
@@ -150,6 +150,8 @@ class ExtractStatistics:
 
     def yield_files(self, file_to_check='statistics.log'):
         pattern = re.compile('[0-9]+')
+
+        # DAS structure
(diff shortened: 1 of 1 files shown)


devos50:
Whoops, forgot about these PRs, sorry for that!

qstokkink:
@devos50 I left you #321, I just needed these 2 merged 👍

PR #329, opened 2017-05-30
Fixed candidate pubkey var

qstokkink:
Previously, nodes shared their public_key variable as the base64 encoded version of the string: `<M2Crypto.EC.EC_pub instance at 0x7fb30213d8c0>`. So this is not actually the public key.


qstokkink:
@devos50 ready for review!

devos50 (approved):

PR #340, opened 2017-12-14
Fail unittests on exit status

qstokkink:
If the process guards exits with status `139` also fail the unit tests.

Previously this would not happen on platforms other than Linux.


qstokkink:
This PR does not fix the issue.
The exit state is correctly returned here (from `wrap_in_temp_home.sh`):

https://github.com/Tribler/gumby/blob/21c6f1bb0700170ba250749068490e2c07ec46b2/scripts/run_nosetests_for_jenkins.sh#L154

However, it is never checked to be 0.

devos50 (approved):

PR #344, opened 2018-01-02
Fixed DispersyModule wrong config function

### LINKED ISSUE #342: Tracker output bleeds into Gumby output
When running a local (Jenkins not confirmed) Gumby job for the first time on a machine, the output of `run_tracker.sh` is shown in the console. This tracker output should be piped to a file instead.

### LINKED ISSUE #343: base_dispersy_module broken
`base_dispersy_module.py` uses `config.set_mainline_dht_listen_port()` instead of `config.set_mainline_dht_port()`, this crashes Gumby.

qstokkink:
Fixes #342

Fixes #343


devos50 (approved):
As a side note, maybe we should specify somewhere that Gumby is compatible with the `devel` branch of Tribler?

qstokkink:
Good idea 👍

PR #348, opened 2018-01-03
Fix for run_nosetest  being a string

### LINKED ISSUE #347: run_nosetests_for_jenkins.sh broken
The latest change (dd286713f0d36c545ab5198e7fd8b08977781c98) broke https://github.com/Tribler/gumby/blob/devel/scripts/run_nosetests_for_jenkins.sh for Windows and Mac.

For some reason this line:
https://github.com/Tribler/gumby/blob/dd286713f0d36c545ab5198e7fd8b08977781c98/scripts/run_nosetests_for_jenkins.sh#L160-L161
Does not properly split the space delimited `$pids` into different `p`, but instead outputs:
```
line 161: wait: ` 24496 24498 24500 24502 24504': not a pid or valid job spec
```

qstokkink:
Fixes #347 

For some reason `$pids` is being interpreted as a delimited string on the Windows and Mac machines. This PR makes it so that if either `$pids` is equal to `  1 2 3 4` (as it should be) or `" 1 2 3 4"` (as it apparently is on the Windows and Mac machines), the for loop will work.

### DIFF
--- a/scripts/run_nosetests_for_jenkins.sh
+++ b/scripts/run_nosetests_for_jenkins.sh
@@ -157,7 +157,7 @@ else
                 pids+=" $!"
                 let COUNT=1+$COUNT
             done
-            for p in $pids; do
+            for p in `echo $pids`; do
                 if wait $p; then
                     echo "Process $p success"
                 else


qstokkink:
@devos50 ready for review. This one is kind of critical, as all Tribler PR unit tests for Windows and Mac will fail due to this.

devos50 (approved):

devos50:
Is it really that hard to make a compatible bash interpreter across all three major platforms? :/

PR #352, opened 2018-01-08
WIP: Fix gumby with new tribler config structure

Captain-Coder:
In this pull request:
- Refactored the code that accesses dispersy(provider)/session/config to be in BaseDispersyProvider. Other experiment modules also need access to these items, and there was some duplicated code as a result of this.
- Renamed tribler_config to session_config, this should now be consistent across all gumby modules.
- Changed the CommunityExperimentModule to return the correct tribler config (instead of the old tribler-session-is-also-a-config)
- Renamed Multichain references to TriblerChain


qstokkink:
Ok to test

devos50:
I guess the issues this PR aims to fix, are already fixed by now 👍

PR #353, opened 2018-01-18
WIP: Match gumby experiment client id's to process guard's log numbers

Captain-Coder:
I got fed up with trying to work out which log file corresponds to what experiment client. 

This PR changes the experiment server to keep a map of what IDs clients have requested. Clients that have not requested an ID get assigned one when the IDs need to be sent. Furthermore, the process guard sets an environment variable that communicates to the experiment client what its "number" is. The experiment client then uses this number to request that specific ID from the experiment server.

Note that when using more than one process guard to launch experiment clients (like on the das), all guarantees go out the window. This could be fixed in the future by adding an offset through the environment of the process guard such that each process guard sets up a disjoint block of ids.

### DIFF
--- a/gumby/sync.py
+++ b/gumby/sync.py
@@ -78,10 +78,10 @@ class ExperimentServiceProto(LineReceiver):
     # Allow for 4MB long lines (for the json stuff)
     MAX_LENGTH = 2 ** 22
 
-    def __init__(self, factory, id):
+    def __init__(self, factory):
         self._logger = logging.getLogger(self.__class__.__name__)
 
-        self.id = id
+        self.id = None
         self.factory = factory
         self.ready = False
         self.state = 'init'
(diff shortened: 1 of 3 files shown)


qstokkink:
Ok to test

Captain-Coder:
Poke @devos50, @qstokkink or @xoriole to review & merge this.

gumby/sync.py
```diff
@@ -193,7 +209,8 @@ def setConnectionMade(self, proto):
             if self._made_looping_call and self._made_looping_call.running:
                 self._made_looping_call.stop()
 
-            self.pushIdToSubscribers()
+            # this gives the reactor some seconds to process pending id request messages
+            reactor.callLater(5 + 0.2*self.expected_subscribers, self.pushIdToSubscribers)
```
devos50, line 213: What happens if we run an experiment with 1000 nodes? Then I would expect the `self.expected_subscribers` array to contain 1000 elements, so I have to wait at most 205 seconds?

Captain-Coder: Yes. Perhaps we can taper this slightly to increase the base 5 to 10 and reduce the 0.2 to 0.05? (== 60 seconds for 1000 nodes) But I didn't have the scale to test this properly.

Since I implemented requesting an ID as an optional feature, I couldn't call pushIdToSubscribers() depend on messages that might never come. I tried to find something like "if reactor is idle 5 seconds then invoke deferred", but I couldn't find it. (Not that the twistd docs are that great for finding what you are looking for)

However as I said in the OP, when running this on the DAS, a little more work is needed. Each process_guard will allocate the same block of ID's. So I wasn't really thinking about the 1000 nodes scenario. (If that really exists, we might want to check the default port allocations on gumby too. That is set up with max a few hundred in mind currently)

devos50:
Although this feature is very helpful to have (especially if running 1000+ instances), we require some more discussion on how to exactly achieve this without delaying the experiment duration too much. For now, I will close this issue due to inactivity.

PR #358, opened 2018-02-05
Set peer id during debug

Captain-Coder:
When debugging in an IDE, a run of a gumby scenario assigns experiment client ID 1. However that might not be the ID you want to debug. This PR adds an environment variable to be set by the IDE that forces the experiment runner to set a certain ID number, regardless of what the server assigned it.


devos50:
ok to test

Captain-Coder:
@devos50, @qstokkink or @xoriole , can someone review and merge?

devos50 (approved):

PR #368, opened 2018-08-08
Disabled Dispersy by default in Gumby experiments

devos50:
So it doesn't slow down experiments if we forgot to disable it...


qstokkink (approved):
At some point we might also want to rename `base_dispersy_module.py` then :)

PR #377, opened 2018-09-23
Various improvements to TrustChain experiments

devos50:
In this PR, I made various improvements to the TrustChain experiments. I also made a few minor QoL changes.
- Each node now writes away the start time of an experiment (this can be used during post-processing).
- Improved existing TrustChain experiment.
- Added crawler experiment that deploys a walking peer to crawl TrustChain blocks.
- Plotting the total number of TrustChain blocks over time.
- Disabled walker in experiments by default.
- Added a memory database (this comes in handy when doing an experiment that requires low-latency queries for blocks).
- Added a few callbacks to `TrustChainModule` for changing various settings.
- Made it possible to dynamically add new walking strategies to an experiment.
- Added an optional limit parameter to the `introduce_peers` callback.

### DIFF
--- a/experiments/trustchain/trustchain_mem_db.py
+++ b/experiments/trustchain/trustchain_mem_db.py
@@ -1,0 +1,127 @@
+from Tribler.pyipv8.ipv8.attestation.trustchain.block import TrustChainBlock
+
+
+class TrustchainMemoryDatabase(object):
+    """
+    This class defines an optimized memory database for TrustChain.
+    """
+
+    def __init__(self, working_directory, db_name):
+        self.working_directory = working_directory
+        self.db_name = db_name
+        self.block_cache = {}
+        self.linked_block_cache = {}
+        self.block_types = {}
+        self.latest_blocks = {}
+        self.original_db = None
+
+    def get_block_class(self, block_type):
+        """
+        Get the block class for a specific block type.
+        """
+        if block_type not in self.block_types:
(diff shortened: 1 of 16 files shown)


experiments/trustchain/trustchain_mem_db.py
```diff
@@ -0,0 +1,127 @@
+from Tribler.pyipv8.ipv8.attestation.trustchain.block import TrustChainBlock
+
+
+class TrustchainMemoryDatabase(object):
```
qstokkink, line 4: Why not just use the `u":memory:"` working directory?

devos50: Valid question. When doing experiments with the market, I found out that even a sqlite memory database is not efficient enough when trading with a very high volume. If I recall correctly, it had quite a high overhead when validating TrustChain blocks and had a major impact on my results (i.e. the horizontal scalability of the market was not there).

This custom memory database is optimised for block validation and quick lookup blocks, including lookups of linked blocks. It is completely optional to use, which is the reason I have it in Gumby. Also, it is disabled by default so it does not influence the current validation experiments. I'm pretty sure that I will use this database in the near future, probably to stress the the market and/or TrustChain.

qstokkink: Alright.

qstokkink (approved):

PR #390, opened 2018-12-17
READY: Added local DHT storage measurement for DHT find/store based experiments.

DanGraur:
Augmented the DHT tests with a new graph and metric which keeps track of the DHT local storage size, as new stores and finds are performed.


DanGraur:
I should mention the following: in this current version, an experiment node will measure the size of its local DHT storage whenever one invokes the `find()` or `store()` method in the `DHTModule` class. This means that one is dependent on calls to these methods in order to obtain an accurate and up to date view of the DHT's size. I've also thought of an alternative, wherein one uses a `LoopingCall` to the DHT storage size measurement method. This will essentially make the measurement process independent of calls to the `find` and `store` methods, but I wasn't sure if it's a desired behavior. I can also wrap the `start` method of the `LoopingCall` in a `@experiment_callback` method, such that the author of an experiment can choose to measure the DHT size during the experiment (instead of having it on by default, say either in the `__init__` or `on_id_received` methods). Please let me know if I should implement this.

A sample graph obtained via the code I wrote for this can be seen here (ran on 1 machine with 10 processes):

Alt Text

I've also enabled annotations in DHT experiments, as well as autoplot aggregation. I've done this using delegation to an `IPv8StatisticsParser`. Please let me know if this is not ok. Also, on a related topic, I've removed the annotations from the .r script plotting the DHT `find` and `store` response times (that is, `dht.r`), since they were not suitable for the response time plot.

synctext:
add to whitelist

DanGraur:
I added what I proposed earlier. i.e. a `LoopingCall` which performs the measurements independently of the `find` and `store` calls. The looping call needs to be triggered by calling an `experiment_callback` (in this case the `trigger_measure_dht_size` method). The `LoopingCall` samples the size of the local DHT storage every 0.2 seconds by default (but this can be modified by passing a parameter to the callback directly in the `.scenario` file). This is an example graph generated by the test (test run on 1 machine with 10 processes):

devos50:
Ok to test

devos50:
@synctext you are probably not an admin for the Gumby tester on Jenkins. I will add you to this list.

devos50:
@DanGraur I would say that deciding whether to take a storage size sample with fixed time intervals or when the storage itself is changed, depends on the frequency of updates. When going for the "dynamic" sampling policy you should make sure that the overhead of the sample does not influence your results when the DHT is under pressure. In such a scenario, I would sample the DHT every x seconds, where x depends on your experiment duration.

DanGraur:
@devos50 understood. I'll make the necessary changes. Also, knowing that I can write some custom code in the other applications such as IPv8, to help with testing, do you think it would be a better idea if I have several callbacks which get called when an element is removed / added, and decrements / increments the running size of the local DHT storage (which should be much faster, but more tricky to implement - not that that should be a problem, though -), rather than a method which is called periodically (even if the call interval is relative to the running time of the test)?

qstokkink:
In general we tend to avoid adding code to a project for testing purposes. If proper design patterns are followed, the code should be naturally testable.

DanGraur:
I've changed the `LoopingCall` such that it now depends on the experiment time and the number of samples required. The programmer will now have to manually pass the experiment time to the `trigger_measure_dht_size`, and optionally the number of samples required (this is by default `25`). The LoopingCall is still there, and will now call the measurement metric every `float(experiment_time) / float(sample_count)` seconds. If the passed experiment time is accurate, then the measurement method will be called `sample_count` times. I think this can be improved, since it relies too heavily on the programmer specifying the experiment time, which they might get wrong. I'll try to find a way to get the experiment time programatically, directly from the experiment module.

DanGraur:
As per @qstokkink's comment, I did not add any custom code (in pyipv8 or Tribler) for this experiment. The measurement still relies on a `LoopingCall`. To ensure that the looping call does not go on for too long, it will automatically be terminated after the specified `experiment_time` (although this happens, using the aforementioned failsafe, only when the `experiment_time` is actually shorter than the moment `stop_session` is called; the `LoopingCall` is killed by default otherwise). I also squashed the commits and rebased the branch. Finally, I'm attaching two examples here for the obtained graphs with a more reduced number of samples.

25 samples:



10 samples:

devos50:
@DanGraur thanks for your changes.

> In such a scenario, I would sample the DHT every x seconds, where x depends on your experiment duration.

What I intended to say here is that the value of x should be sane and not too big or too small, but it does have to be directly dependent on the total experiment duration. The problem here is that the total experiment duration is hard to get (we probably need to inspect the actions in the scenario file for this). Usually, when I write experiments, I trigger the looping call with a fixed time interval. For example, see [here](https://github.com/Tribler/gumby/blob/devel/experiments/trustchain/trustchain_module.py#L64). I would suggest modifying your code so you avoid using the experiment duration in your code. For example, you can make the `sample_count` the only parameter in `trigger_measure_dht_size`, which is defined by the programmer. The programmer should then make sure that this sample count makes sense with respect to the total experiment duration.

DanGraur:
@devos50 ok, I think there was a misunderstanding here. In a sense, I already implemented it like that initially: in my initial implementation the `trigger_measure_dht_size` method had a singular parameter which would control the interval between samples. This was by default set to `0.2`, but as I explained in [one of my previous comments](https://github.com/Tribler/gumby/pull/390#issuecomment-447905884) this could be changed by the writer of the `.scenario` file directly in the scenario by passing a single parameter which represented the sampling period, thus overriding the default value of `0.2`. So to clarify, the parameter represented the **sampling period**, not the **sampling count**. That being said, I think the  **sampling count** is not that useful if you don't have the experiment time, since you cannot derive a sampling period (this is essentially the current implementation, where the programmer has to specify the duration of the experiment, and optionally the sampling count). The **sampling period** makes sense in my opinion, though, since the programmer should be able to infer it with relative ease (since they can also judge the general DHT load during the experiment as well as easily tell the duration of the experiment). I'll reimplement my initial implementation, since I can't revert to that commit (it was removed when tidying up the commit history). It should take little time, though, since there's little code involved in the change 🙂.

synctext:
Dan, impressive detailed work! Is there an issue and formulated goal for this test? I'm a bit lost, like stress testing at scale.

DanGraur:
I created a ticket for this PR (#393). I hope it describes the issue at hand fairly well, but if there's anything I can do to improve the ticket please let me know. I'll be more than glad to help!

DanGraur:
I've added a large scale Jenkins job for this (100 nodes): https://jenkins-ci.tribler.org/job/dht_local_size_experiment/. I've also moved this experiment to its own scenario and config files (both local and on the DAS). I think this PR is ready now.

qstokkink:
Cool. I'll give it a review.

@devos50 do we want those 2 experiments to be part of the validation experiments?

experiments/dht/dht_module.py
```diff
@@ -17,16 +21,45 @@ def __init__(self, experiment):
         super(DHTModule, self).__init__(experiment, DHTDiscoveryCommunity)
         self.start_time = 0
 
+        self.callbacks = {}
+
     def on_id_received(self):
         super(DHTModule, self).on_id_received()
         self.tribler_config.set_dht_enabled(True)
 
+        with open('autoplot.txt', 'a') as output_file:
```
qstokkink: Have a look at https://github.com/Tribler/gumby/blob/e8c77759bc6cfbffb75a272e92c98ed4ced1eb4f/experiments/gigachannel/gigachannel_module.py

There is no need to make the autoplot directories yourself.

experiments/dht/dht_module.py
```diff
@@ -51,6 +84,22 @@ def on_peer_store_error(failure):
...
+        Compute the size of the storage in the DHTOverlay
+
+        :return: None
+        """
+        time_stamp = time.time()
+        total_storage_size = 0
+
+        for key in self.overlay.storage.items:
+            for item in self.overlay.storage.get(key):
+                total_storage_size += getsizeof(item)
```
qstokkink: Just `sum()` should be just fine here.

qstokkink (changes requested):
Looking good, please fix these 2 things.

DanGraur:
Should be good now, I think.

experiments/dht/dht_module.py
DanGraur: Was [this](https://github.com/DanGraur/gumby/blob/d5f6116440be4d1420fe74a90b204150ad519dc4/experiments/dht/dht_module.py#L88) what you had in mind for the `sum()`?

devos50:
@DanGraur what is the status of this PR?

DanGraur:
@devos50 , I've taken a look at the PR's code, and it all seems fine. All the changes requested in this thread were implemented. This PR can be merged after the conflicts are resolved and the branch is rebased. I'll get on this.

devos50:
I'm not sure if this will be merged on short term. Will re-open if we work on additional DHT experiments 👍

PR #397, opened 2019-02-11
Variables in Scenario files.

DanGraur:
Added the possibility of declaring variables within scenario files. Variables can now be added by using: `@! set <var_name> <var_value>` anywhere within the scenario file. A variable's scope will begin from the moment it was declared, until the end of the file. The variable can then be used as an argument in other function calls using: `$<var_name>` (e.g. `@0:10.28 store 10d00d55231921911991e2f7fd538ec989a2df00 $val {7}`, where `val` was declared using `@! set val value_to_store`). 

I've also added a new construct, which might need to be discussed and agreed upon: `@!`. The implementation relies on this particular element. It essentially means 'execute this procedure now'. Currently, when the `.scenario` file is parsed, each line is read, then split into tokens, then scheduled for execution at a particular moment in time using Twisted. `@!` circumvents this, and essentially tells the parser 'do not use Twisted to schedule this in the future, execute this instruction now', i.e. now during the parsing phase, which is before the actual experiment phase. This is needed since when we use something like `@0:1 set my_variable 10`, Twisted will schedule the execution of this instruction 1 second in the experiment. If we have another instruction like `@0:15 find_in_dictionary $my_variable`, although this is 15 seconds into the experiment, during parsing, when this instruction is scheduled, the variable will not be visible, since it was not actually executed yet. Using `@! set my_variable 10` instead, will solve this issue, since the value of the variable `my_variable` will be set during parsing, and as such, `my_variable` will be visible for `@0:15 find_in_dictionary $my_variable` as well.


devos50:
ok to test

devos50:
add to whitelist

devos50:
@DanGraur when done, please prefix the PR title with READY 👍

qstokkink (approved):

devos50 (approved):

PR #398, opened 2019-02-18
Simplified code for scenario variables 

DanGraur:
After taking a second look at the code that I wrote for adding variables to a `.scenario` file, I realized that there's an easier, and more organic way of making use of them in the `.scenario` files. I removed the cumbersome code from `_parse_scenario_line` (which made explicit checks for the presence of variables to then replace them with their assigned value) and added a few very simple lines in `_expand_line`, which was already looking for variables, i.e. tokens starting with the `$` character. `_expand_line` was essentially looking for environment variables, now I've extended its behavior to look for user defined variables. In terms of precedence, the user defined variables are preferred to the environment variables. I can change that if you'd like. Furthermore, as a result of this change, when the variable name is not found in either the set of user defined variables and the set of environment variables, it will just be treated as a literal (with `$` included in its name). With my previous implementation, when this happened, an exception was raised. I'm not sure what is best here, so if you'd like me to raise an exception, please let me know (it should be an easy change to make).

### DIFF
--- a/gumby/scenario.py
+++ b/gumby/scenario.py
@@ -196,15 +196,6 @@ class ScenarioParser(object):
                     if argname:
                         named_args[argname.group(1)] = arg[argname.start(2):]
                     else:
-                        # If this is a variable
-                        if arg[0] == '$':
-                            # Get the name of the variable
-                            var_name = arg[1:]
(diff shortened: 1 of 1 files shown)


qstokkink (approved):
I think it's fine as is.

devos50 (approved):

PR #399, opened 2019-02-21
Stripped Dispersy from Gumby

devos50:


docs/hello_world.rst
```diff
@@ -1,246 +0,0 @@
-The purpose of this document is to showcase creation of a basic Gumby experiment.
```
qstokkink: Why was my beautiful documentation removed? 😢

devos50: Because it is completely based on Dispersy. I am planning on rewriting it and I should probably do that tomorrow.

qstokkink: 98% of this file is about general Gumby configuration, throwing it all out seems like a waste.

devos50: You're right, I didn't read through it properly. I made a few changes to make the last section use IPv8 instead.

devos50:
For some reason, libtorrent is not working correctly in a newly setup DAS5 environment. I will look into it and fix this in another PR. The current DAS5 environments are working correctly and as expected.

qstokkink (approved):

PR #401, opened 2019-02-22
Improved Gigachannels experiment

devos50:
In this PR, I improved the gigachannels experiment so it runs on the DAS5 and is compatible with the latest code.

Also, I'm plotting the total number of torrent entries in the database now, which looks something like this:

total_torrents


devos50:
@qstokkink ready for review. I will add channel subscription in another PR.

qstokkink (approved):
Cool. Remind me monday to make autoplot a bit easier to work with.

PR #406, opened 2019-02-27
Made procfs optional

### LINKED ISSUE #253: Running Gumby on OS X and Windows
In order to be able to run Gumby on OS X and Windows (for the unit tests of Tribler/Dispersy), some major changes are needed: 

`procfs` is not available on OS X and the port (https://github.com/osxfuse/filesystems/tree/master/filesystems-c/procfs) seems to be outdated. On Windows, we have no real good alternative for `procfs`. `psutil` is providing some tools to get statistics. So we might fallback on `psutil` if we do not have `procfs`.

Another issue is that Windows does not support process groups.

This issue is open for discussions, ideas and suggestions how to make Gumby cross-platform.

devos50:
If procfs is not available, we return zeros. This makes it possible to run Gumby on Mac and/or Windows.

The procfs stats on Linux are still working, as indicated by the integration test that runs as part of this PR.

Fixes #253

### DIFF
--- a/scripts/process_guard.py
+++ b/scripts/process_guard.py
@@ -5,6 +5,7 @@
 from __future__ import print_function
 import exceptions
 import json
+import os
 from glob import iglob
 from math import ceil
 from os import (R_OK, access, errno, getpgid, getpid, kill, killpg, makedirs, mkdir, path, setsid, sysconf,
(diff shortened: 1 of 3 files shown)


scripts/process_guard.py
```diff
@@ -96,47 +97,45 @@ def pid_exists(pid):
...
-                        "wchar: 0\n",
-                        "syscr: 7\n",
-                        "syscw: 0\n",
-                        "read_bytes: 0\n",
-                        "write_bytes: 0\n",
-                        "cancelled_write_bytes: 0\n",
-                    ]
-                else:
+            if not os.path.exists('/proc'):
+                dummy_status = "%s (sh) S 0 0 0 0 -1 8192 0 0 0 0 0 0 0 0 0 0 1 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 " \
```
xoriole, line 101: What is the total expected size of records here?

qstokkink (approved):

scripts/process_guard.py
devos50: 52 items, see the procfs manual: http://man7.org/linux/man-pages/man5/proc.5.html

xoriole (approved):

PR #407, opened 2019-03-03
Added the possibility of using for loops within scenario files.

DanGraur:
It is now possible to use for loops inside scenario files. For loops can also make use of their control variable to either set parameters in the function call or to select which nodes will execute the for loop itself. This PR also adds a few scenarios specifically for exemplifying and testing the behavior of the for loops.

For loops we have the following syntax:
```
@hh:mm:ss for <control_variable> in <left_bound> to <right_bound> call <callable> [<unnamed_params>] [<named_params>] [<peerspec>]
```
The `<left_bound>` and `<right_bound>` are inclusive, and need not be `<left_bound> <= <right_bound>`, i.e. it can also be `<right_bound> >= <left_bound>`. The increments are currently either `+1` or `-1`, depending on the bounds.

It is possible to use the `<control_variable>` as a named or unnamed parameter. It should be noted, however, that if there exists a variable with the same name, the variable takes higher precedence as a parameter, and its value will be used. Should I change this behavior such that an error is thrown when we have conflicting names?

It is also possible to use the `<control_variable>`, or a scalar in the `peerspec` part of the `for` loop (not both at the same time). It is not possible to use exclusion semantics together with the `<control_variable>` yet, i.e. use `!` before `$<control_variable>` in the `peerspec` to denote that a node should not invoke the callable for this iteration.

### DIFF
--- a/docs/advanced_scenario_concepts.rst
+++ b/docs/advanced_scenario_concepts.rst
@@ -1,0 +1,237 @@
+***********************************
+Advanced Scenario Language Concepts
+***********************************
+
+The document is meant to exemplify some of the more advanced features of the scenario language. Currently, these features refer to:
+
+- Support for ``variables``
+- Support for ``for`` loops
+
+They will be exemplified and presented in further detail in the sections to follow:
+
+Variables
+---------
+
+Variables have been introduced in the scenario language mainly for those cases when a value is repeatedly used during an experiment. Without variables this would force a programmer to frequently rewrite or copy-paste the literal, which may introduce human-made errors, and is at the very least cumbersome and tiring. Variables can help with this issue, by associating a value to a name / alias.
+
+Variables are by default **strings**, and it is the responsibility of the programmer to ensure that they convert the string value to whatever type they require in the function code. This, however, does not differ to how normal literal parameters need to be processed within a function, as they are **strings** by default as well.
+
+The syntax for declaring a variable is the following:
+
+``@! set <variable_name> <variable_value>``
+
+**The scope of the variable begins immediately and persists until the end of the scenario file**. The meaning of each component is explained in what follows:
+
+- ``@!`` is a special timestamp which informs the parser that the command must be executed immediately, as opposed to using Twisted to schedule it in the future, after the scenario has been parsed and the experiment has begun.
+- ``set`` is a special experiment callback, which is used to declare variables
+- ``<variable_name>`` is the name of the variable
+- ``<variable_value>`` is the value assigned to the variable
+
+One can redefine a variable multiple times, and **the scope of the new value begins immediately, while the scope of the previous value will end immediately as well**.
+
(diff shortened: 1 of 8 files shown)


qstokkink (changes requested):
Nice 👍

Could you add some `# TODO:`'s to the regex and peer parse sections to show future programmers where to implement `peerspec` substitutions?

devos50 (commented):
Nice addition!

I have a general comment. I noticed that you did add some new experiments to Gumby (which executes a scenario file with some for loops in it). I would prefer that these files are actively used in either an experiment on Jenkins (e.g. as part of the testers when making a PR) or within a unit test. I think that running this experiment as an experiment itself is a bit too much and it would be nice if you could create a unit test out of it. We already have a few unit tests that are executed when making a PR (see the `gumby/tests` directory).

What do you think about this?

DanGraur:
Sorry for the late response @qstokkink and @devos50. Will do on both these requests!

DanGraur:
I've finished adding a test class which is automatically run using Jenkins. The class is called `TestForLoop` and is located in the `tests` package. It features unit tests which should validate the correctness of the `for` loop implementation, as well as exemplify how it can be used. I've also added two `TODO`s which should detail how to further process a line in the scenario file, such that the code can be simplified, but actually implementing this might require some effort.  

I've also left in the scenarios which demonstrate how to use the `for` loops, as well as the class `ForLoopTestModule`. Should I remove them, or can they stay in?

qstokkink:
> I've also left in the scenarios which demonstrate how to use the for loops, as well as the class ForLoopTestModule. Should I remove them, or can they stay in?

Please keep them. Our documentation is already pretty sparse. The more documentation we have, the better.

Maybe it's best to move them into a single `.rst` (maybe `forloops.rst`?) file in the `/doc` folder though, otherwise they might get thrown out.

DanGraur:
Ok, I've added the `.rst` file in the `/doc` folder. If it's ok, I called it `advanced_scenario_concepts.rst` (even though they might not be that advanced) since it also documents variables, not just `for` loops. I should mention that I also left the scenarios which exemplify the `for` loop and its limitations (these two scenarios are in `experiments/test_scenarios/for_loop`), and also a helper class called `ForLoopTestModule` (also located in `experiments/test_scenarios/for_loop`) whose methods are used as `experiment_callback`s  in these aforementioned scenarios. Should I leave those in, or should I remove them, now that there is some documentation for these new features?

If need be, I can remove them, as I have a local commit where I've done this (so I just need to push it). The code won't be usable though, as it relies on the `ForLoopTestModule`, and also requires the `.conf` files. Still, if I delete them, it should be easy for someone else to write a quick `.conf` file, and a helper test class if they wish to run the code for themselves.

qstokkink (approved):
Nice.

@devos50 any comments on this?

devos50 (approved):

devos50:
Nice contribution!

PR #412, opened 2019-04-21
Make sure not to connect all clients at once

devos50:
When spawning an experiment client, it immediately connects to the experiment server. This could overload the experiment server when running > 2000 clients, leading to TCP failures. This in turn occassionally fails the experiment.


qstokkink (approved):
I think it would be better to have some sort of retry system, but this will do for now.

PR #413, opened 2019-05-22
WIP: Added AnyDex isolated module + various small improvements

devos50:
In this PR, I added a separate AnyDex module which loads its own IPv8 instance and initializes various overlays. I also made a few other changes:
- From now on, we don't assume that the IPv8 module is always inside a `Tribler` directory. Instead, you can now optionally pass a `IPV8_DIR` env variable with the path to the `pyipv8` folder. This increases flexibility and is a first step towards decoupling Tribler from IPv8 in Gumby.
- Added a script to collect generated Yappi profile logs and move them to a separate directory.
- Extracted the specific community loaders from the community loader class itself.


devos50:
retest this please

PR #417, opened 2019-06-15
Fixed TrustChain memory DB bug

devos50:
Fixed a bug that caused the memory-based implementation of the TrustChain database to fetch the wrong blocks during a crawl.


qstokkink:
I don't really get this change:

```python
>>> for ind in xrange(1, 3 + 1):
...     print ind + 1
...
2
3
4
>>> for ind in xrange(2, 3 + 2):
...     print ind
...
2
3
4
```

Seems functionally equivalent?

devos50:
Oh whoops, the begin index should be `1` instead of `2`.

Your code is indeed equivalent in functionality but the 'old' code had the wrong return value (line 82).

qstokkink (approved):

PR #441, opened 2019-11-22
Assigning peer ID based on host IP

devos50:
Currently, peer IDs when running an experiment on the DAS5 are assigned on a first-come-first-serve basis. This makes peer assignment a very unpredictable and unreliable process, causing us to manually search amongst all node output for the artifacts of a peer with a specific ID.

This PR changes this behavior and assigns the peer IDs based on the host IP. This means that if you run an experiment on two DAS5 nodes with 16 instances in total, peer 1-8 will be assigned to instances running one host, and peer 9-16 will be assigned to instances on the other host. I changed the dummy experiment to allocate two nodes instead and verified its correctness.

### DIFF
--- a/gumby/sync.py
+++ b/gumby/sync.py
@@ -193,6 +193,19 @@ class ExperimentServiceFactory(Factory):
             if self._made_looping_call and self._made_looping_call.running:
                 self._made_looping_call.stop()
 
+            # Assign IDs to the connected subscribers, based on their address
+            def split_ip(ip):
+                return tuple(int(part) for part in ip.split('.'))
+
+            def ip_addr_key(connection):
+                return split_ip(connection.transport.getPeer().host)
+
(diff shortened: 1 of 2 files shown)


gumby/sync.py
```diff
@@ -193,6 +193,19 @@ def setConnectionMade(self, proto):
...
 
+            # Assign IDs to the connected subscribers, based on their address
+            def split_ip(ip):
+                return tuple(int(part) for part in ip.split('.'))
+
+            def ip_addr_key(connection):
+                return split_ip(connection.transport.getPeer().host)
+
+            self.connections_made = sorted(self.connections_made, key=ip_addr_key)
+            peer_index = 1
```
qstokkink, line 204: You could replace this with:

```python
for i, connection in enumerate(self.connections_made):
    connection.id = i + 1
```

Then again, this is also readable.

qstokkink (approved):

PR #445, opened 2020-01-28
Moved Gumby to asyncio

devos50:
Most of the credits go to @egbertbouman, I just took over the branch and made some additional changes.

This PR has been thoroughly tested with *all* validation experiments, and [our DLT benchmarks](https://jenkins-ci.tribler.org/job/dlt_benchmarks/) that I ported to use `asyncio`. I think it is stable enough to be reviewed and merged 👍


egbertbouman:
When porting Gumby to asyncio, I rewrote the `runRemoteCMD`, and substituted our custom Twisted SSH stuff with asyncssh. However, I never tested this. In fact, none of our testers/experiments seem rely on this logic. Maybe we could even remove it entirely?

qstokkink (approved):
Probably fine?

PR #454, opened 2020-04-10
Added blockchain experiments

devos50:
This PR adds various blockchain experiments to Gumby for the following platforms:
- Algorand
- BitShares
- Ethereum
- Hyperledger Fabric
- Libra
- Stellar

It also changes the synchronization server such that it can be used as a message broker (first commit). This is, for example, needed to have instances share private keys of wallets with each other.


devos50:
This PR is ready for review. For some reason, flake8 does not recognise `requests` as part of the Python 3 standard library, so please ignore the import errors in the Pylint report. Regarding the pylint report, I tried to clean up as many violations as I could, however, some of them would require extensive re-engineering.

qstokkink:
Ah yes 30 000 lines, I'm sure I will have this reviewed by the end of the year.

From a superficial review it seems that everything is compartmentalized correctly and this PR doesn't break anything; I would be fine with just merging it without an in-depth review.

If you want me to do a serious in-depth review, it will take a lot longer.

devos50:
Note that most of the line changes are actually from included scenario files. Still, I think that a superficial review is sufficient here, since the code is not critical for our continuous operation and I won't want you to review everything in-depth (after all, all the experiments pass in Jenkins and a review would require in-depth knowledge of some o the included systems). 👍

devos50:
@qstokkink so, please approve :D

grimadas (approved):

PR #460, opened 2020-08-07
Various fixes

devos50:
This PR fixes various issues related to the `AnydexModule` class, which loads Gumby without Tribler.

### DIFF
--- a/experiments/market/market_experiment_simple.scenario
+++ b/experiments/market/market_experiment_simple.scenario
@@ -10,6 +10,7 @@
 @0:5 init_matchmakers
 @0:5 set_validation_range 0
 @0:6 init_trader_lookup_table
+@0:7 introduce_peers
 @0:10 connect_matchmakers 10
 @0:25 annotate start-trading
 @0:25.1 bid 10 DUM1 10 DUM2 {1}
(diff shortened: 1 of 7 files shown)


gumby/modules/tribler_module.py
```diff
@@ -262,8 +262,8 @@ def write_overlay_statistics(self):
 
         # Write bandwidth statistics
         with open('bandwidth.txt', 'w') as bandwidth_file:
-            bandwidth_file.write("%d,%d" % (self.session.ipv8.endpoint.bytes_up,
-                                            self.session.ipv8.endpoint.bytes_down))
+            bandwidth_file.write("%d,%d" % (self.session.ipv8.endpoint.interfaces["UDPIPv4"].bytes_up,
```
qstokkink: Hmm.. this is not pretty. I'll add the `bytes_up` and `bytes_down` properties to the `DispatcherEndpoint`.

qstokkink: If you want to wait for https://github.com/Tribler/py-ipv8/pull/837, you don't need this change.

qstokkink: If you use the latest IPv8 you can 
~~drop commit d6df84dbca2d65f74fd9a879e3d46b9bf692f28d from this PR~~, or rather edit it, it seems -- that contains a bit exta.

devos50: Fixed, thanks for the addition! 👍

qstokkink (approved):

gumby/modules/tribler_module.py
qstokkink: If at all possible I'd like IPv8 to good looking, complete and easy to use 😄

PR #468, opened 2020-10-13
Removed unused settings + memory profiler

devos50:
Our memory profiler does not work on Python 3 so I removed it.


qstokkink:
What about this?

https://github.com/Tribler/gumby/blob/1594055d9c147cd8f6c4867562ad67d0995d8e0b/scripts/run_tracker.sh#L79-L82

qstokkink (approved):

PR #470, opened 2020-10-13
Changed default state directory

devos50:
This state directory is by default outside the current working directory, complicating scripts that actually have to search for stuff in the state directory. This commit moves the default state directory to the current working directory, making the experiment output more organized and easier to work with.

### DIFF
--- a/gumby/modules/base_ipv8_module.py
+++ b/gumby/modules/base_ipv8_module.py
@@ -1,8 +1,8 @@
 import json
+import os
 import time
 from asyncio import Future
 from binascii import hexlify
-from os import environ, makedirs, symlink, path, getpid
 from socket import gethostbyname
 
 from gumby.experiment import experiment_callback
@@ -30,7 +30,7 @@ class BaseIPv8Module(ExperimentModule):
         self.session = None
         self.tribler_config = None
         self.ipv8_port = None
-        self.session_id = environ['SYNC_HOST'] + environ['SYNC_PORT']
+        self.session_id = os.environ['SYNC_HOST'] + os.environ['SYNC_PORT']
         self.custom_ipv8_community_loader = self.create_ipv8_community_loader()
         self.ipv8_available = Future()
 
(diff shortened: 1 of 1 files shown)


gumby/modules/base_ipv8_module.py
```diff
@@ -96,14 +96,12 @@ def setup_config(self):
...
-        head_host = gethostbyname(environ['HEAD_HOST']) if 'HEAD_HOST' in environ else "127.0.0.1"
+        head_host = gethostbyname(os.environ['HEAD_HOST']) if 'HEAD_HOST' in os.environ else "127.0.0.1"
 
-        my_state_path = path.abspath(path.join(environ["OUTPUT_DIR"], ".%s-%d-%d" % (self.__class__.__name__,
-                                                                                     getpid(), self.my_id)))
-        makedirs(my_state_path)
-        bootstrap_file = path.join(environ['OUTPUT_DIR'], 'bootstraptribler.txt')
-        if path.exists(bootstrap_file):
-            symlink(bootstrap_file, path.join(my_state_path, 'bootstraptribler.txt'))
+        my_state_path = os.getcwd()
```
qstokkink: `os.getcwd()` is not a stable path, it can change at any moment by anyone and anything.

I'd prefer anchoring folders from stable points like `os.environ['OUTPUT_DIR']` or even the top-level `run.__file__` file.

qstokkink (changes requested):

gumby/modules/base_ipv8_module.py
devos50: That we change current working directory is actually a feature of Tribler. Specifically, the moment that we receive a peer ID from the central experiment server, we change the current working directory to a directory with the peer ID in there.

I'll try to think of a different solution for this.

devos50: I have implemented an alternative solution, where I explicitly set the right directory.

qstokkink (approved):
Much appreciated 👍

PR #473, opened 2020-10-27
Added basic bandwidth accounting experiment

devos50:
This PR adds a basic bandwidth accounting experiment, following the merge of https://github.com/Tribler/tribler/pull/5626 in Tribler. The experiment initiates random bandwidth payouts with random users, and afterwards aggregates all payouts in a single database.

The validation experiment can be found [here](https://jenkins-ci.tribler.org/job/validation_experiments/job/validation_experiment_payouts/).

### DIFF
--- a/experiments/bandwidth_accounting/bandwidth_accounting_module.py
+++ b/experiments/bandwidth_accounting/bandwidth_accounting_module.py
@@ -1,0 +1,38 @@
+import random
+
+from tribler_core.modules.bandwidth_accounting.community import BandwidthAccountingCommunity
+
+from gumby.experiment import experiment_callback
+from gumby.modules.community_experiment_module import IPv8OverlayExperimentModule
+from gumby.modules.experiment_module import static_module
+from gumby.util import run_task
+
+
+@static_module
+class BandwidthAccountingModule(IPv8OverlayExperimentModule):
+    def __init__(self, experiment):
+        super().__init__(experiment, BandwidthAccountingCommunity)
+        self.payouts_task = None
+
+    @experiment_callback
+    def start_payouts(self):
+        delay = random.random()
+
+        def start():
+            self._logger.info("Starting random payouts!")
+            self.payouts_task = run_task(self.do_random_payout, interval=1.0)
+
+        run_task(start, delay=delay)
+
+    def do_random_payout(self):
+        verified_peers = list(self.overlay.network.verified_peers)
+        random_peer = random.choice(verified_peers)
+        random_amount = random.randint(1 * 1024 * 1024, 10 * 1024 * 1024)
+        self._logger.info("Performing payout to peer %s with amount %d", random_peer, random_amount)
+        self.overlay.do_payout(random_peer, random_amount)
+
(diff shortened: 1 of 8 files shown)


experiments/bandwidth_accounting/bandwidth_accounting_module.py
```diff
@@ -0,0 +1,39 @@
...
+            self.payouts_task = run_task(self.do_random_payout, interval=1.0)
+
+        run_task(start, delay=delay)
+
+    def do_random_payout(self):
+        verified_peers = list(self.overlay.network.verified_peers)
+        random_peer = random.choice(verified_peers)
+        random_amount = random.randint(1 * 1024 * 1024, 10 * 1024 * 1024)
+        self._logger.info("Performing payout to peer %s with amount %d", random_peer, random_amount)
+        ensure_future(self.overlay.do_payout(random_peer, random_amount))
```
qstokkink: Nitpicking: this works, but why not register the future somewhere if you go through the trouble of ensuring it?

experiments/bandwidth_accounting/post_process_bandwidth_accounting.py
```diff
@@ -0,0 +1,51 @@
...
+
+        with open("total_payout_edges.txt", "w") as out_file:
+            out_file.write("%d" % len(transactions))
+
+    def run(self):
+        self.aggregate_databases()
+
+
+# cd to the output directory
+os.chdir(os.environ['OUTPUT_DIR'])
```
qstokkink, line 48: This is very error-prone. I would suggest stripping this and using `with open(os.path.join(os.environ['OUTPUT_DIR'], "total_payout_edges.txt"), "w") as out_file:` on line 40 instead.

I assume that is the only reason this `chdir` is here, please let me know if that is not the case and I will update my review.

qstokkink (changes requested):

experiments/bandwidth_accounting/post_process_bandwidth_accounting.py
devos50: I had to dig a bit deeper into this, but it seems that the argument passed to the init method of `BandwidthStatisticsParser` is `.`. Therefore, this class makes the hidden assumption that the current working directory is in the experiment output directory, which is definitely not a good design. However, this code has been in Gumby for many, many years now. Fixing this is possible but would require refactoring efforts unrelated to this PR. Therefore, I opt to leave it like this for now. 👍

experiments/bandwidth_accounting/bandwidth_accounting_module.py
devos50: True, that was unnecessary. I removed the `ensure_future`.

experiments/bandwidth_accounting/post_process_bandwidth_accounting.py
qstokkink: Seems reasonable, let's defer that to an issue.

qstokkink (approved):

PR #483, opened 2021-04-15
Added/modified blockchain experiments

devos50:
In this PR, I added the changes I made to Gumby when working on the benchmarking paper. I merge them into the main branch since we will have to conduct some follow-up experiments (involving real-world transaction workloads), and I want to work with a solid and finished foundation. The main changes that this PR contains are as follows:
- Added support to run (blockchain) experiments on arbitrary Linux servers.
- Improved Ethereum experiments, added support for the deployment and invocation of arbitrary contracts.
- Added Avalanche to our benchmarking suite.
- Added Hyperledger Burrow to our benchmarking suite.
- Renamed `DAS4_INSTANCES_TO_RUN` variable to `INSTANCES_TO_RUN` (to make it consistent across all deployment plans).
- Renamed `DAS4_NODE_AMOUNT` variable to `NODE_AMOUNT` (to make it consistent across all deployment plans).
- Renamed `DAS4_NODE_TIMEOUT` variable to `NODE_TIMEOUT` (to make it consistent across all deployment plans).

In addition, I made many improvements to existing scripts and made sure that each system works with the latest software version. To test all these changes, I've created a separate Jenkins directory containing various validation experiments, see [here](https://jenkins-ci.tribler.org/job/dlt_benchmarks/job/validation/).

Known shortcomings:
- We could automatically install the `solc` compiler in case it is not found by Burrow/Ethereum.
- Some paths to binaries (e.g., `geth`) are hard-coded. They should be specified from the Gumby experiment since their location is likely differs across environments.


devos50:
@qstokkink even though I requested a review from you, I do not expect you to give a thorough review since this PR is quite large and requires in-depth knowledge of each tested blockchain system 😀. But it would be nice if you can have a quick look to see if I'm missing something obvious 👍

experiments/burrow/burrow_module.py
```diff
@@ -0,0 +1,285 @@
...
+
+    @experiment_callback
+    def generate_config(self):
+        """
+        Generate the initial configuration files.
+        """
+        self._logger.info("Generating Burrow config...")
+
+        # Remove old config directory
+        shutil.rmtree("/home/martijn/burrow_data", ignore_errors=True)
```
qstokkink, line 97: It seems to me like all of these scripts are heavily tied to your ecosystem. At this moment, it seems this code serves to show others what you are running and not to allow others to run what you are running.

For reuse, I'd suggest (1) adding a (short) description to each file to document what software needs to be installed. I'd also suggest (2) to refactor of all the paths that currently point to your personal directory into a variable (i.e., `self.install_dir / "burrow_data"` instead of `"/home/martijn/burrow_data"`).

I'll leave this as a suggestion and not a request for changes. After all, I'm not sure whether you even want others to run this code (on the other hand, why push this code to the shared repo then?). Let me know if this is out of scope and I'll approve the PR as needed.

devos50: Thanks for the suggestions! Indeed, the current code is tightly integrated with our infrastructure but that has always been the case, even with the DAS5 code. Even though I acknowledge it is not a good pattern, my goal is to make Gumby suitable to run these blockchain experiments on our (Surfnet) infrastructure.

My OP might have been a bit misleading then since it does not support running these experiments on arbitrary servers without a specific setup. Many of these blockchain systems require multiple binaries/support for different compilers etc. For now, I try to keep it simple but ideally, we want to automate the setup phase. To some extent, this has been done already (e.g., the generation of genesis files).

My goal with this PR is to provide the experiment code since it gives valuable insights into the setup/operation of the tested systems and how to deploy a private testnet, even though I acknowledge that it is very, very hard to run it in a different environment.

BTW, we can probably move some of these variables to the configuration. However, I would first like to rework the configuration system, so I'm delaying that effort for now.

qstokkink (approved):

PR #508, opened 2022-02-18
Fix Tribler support & validation experiments

kozlovsky:
This PR makes Gumby compatible with the current component architecture of Tribler.

The initialization process of Tribler components is significantly different from the previous initialization logic of IPv8 overlays, so I decided that it is better not to try to mix both types of logic in the same module, but instead to make Tribler-related modules separate. It is possible because most experiments use either Tribler-related or non-Tribler-related modules, and the current experiments that use both are probably obsolete. In the future, if the same functionality should be used both with and without Tribler, we can have two separate modules for that or come up with a new idea of how to combine APIs without unnecessary complications.

So, I introduced a separate base class for Tribler-related modules (`TriblerBasedModule` instead of `IPv8OverlayExperimentModule`) and a separate class for the "main module" that provides IPv8 instance (`TriblerModule` instead of `BaseIPv8Module`).

Jenkins jobs are fixed as well. With this PR, [validation experiments](https://jenkins-ci.tribler.org/job/validation_experiments/) are working except [hiddent seeding experiments](https://jenkins-ci.tribler.org/job/validation_experiments/job/validation_experiment_hidden_seeding/), that also technically works, but does not produce the desired outcome. That requires additional investigations - probably `TriblerTunnelCommunity` needs some fixes that are better to implement in a separate PR.


devos50:
ok to test

devos50:
add to whitelist

devos50:
Thanks for your work!

I just looked at the validation experiments - it looks like the payout experiment output is redirected to `stdout` (see [here](https://jenkins-ci.tribler.org/job/validation_experiments/job/validation_experiment_payouts)). Please restore this so they write output to files instead (you probably used the `GUMBY_DEBUG` option). 👍

kozlovsky:
Fixed!

experiments/dht/dht_module.py
```diff
@@ -54,7 +62,7 @@ async def log_timing(self, coro, op):
             await coro
             self.write_to_log('dht.log', '%d %s %.3f\n', ts, op, time.time() - self.start_time - ts)
         except DHTError:
-            self.write_to_log('dht.log', '%d %s -1', ts, op)
+            self.write_to_log('dht.log', '%d %s -1\n', ts, op)
```
devos50, line 65: Good find!

devos50:
@kozlovsky it also seems that the Gigachannel experiment is not fixed, see [these log entries](https://jenkins-ci.tribler.org/job/validation_experiments/job/validation_experiment_gigachannels/829/artifact/output/localhost/node301/00002.err/*view*/).

However, these experiments were already broken. I'm not sure how much effort it will take to revive them but given that channel creation/dissemination is a key feature of Tribler, making sure that this experiment works correctly is high on my wish list 👍 . Do you want to fix this in this PR?

kozlovsky:
@devos50 yes, it should be fixed. Thanks you for pointing this! I'll try to do it today.

kozlovsky:
Gigachannel experiments [is apparently fixed](https://jenkins-ci.tribler.org/job/validation_experiments/job/validation_experiment_gigachannels/840/) now, so validation experiments except hidden seeding [are working](https://jenkins-ci.tribler.org/job/validation_experiments/job/validation_experiments/1243/).

experiments/tribler_idle_run/tribler_idle_run.py
```diff
@@ -1,49 +1,56 @@
...
+    print("Stopping Tribler idle run")
+    session.shutdown_event.set()
+
+    # Indicates we are shutting down core. With this environment variable set
+    # to 'TRUE', RESTManager will no longer accept any new requests.
+    os.environ['TRIBLER_SHUTTING_DOWN'] = "TRUE"
+
+    await session.shutdown()
+
+    get_event_loop().stop()
```
devos50: **Suggestion:** Maybe we should move this call to `stop()` outside this method? Also, is this necessary (we are already calling `close()`)?

experiments/tribler_idle_run/tribler_idle_run.py
```diff
@@ -1,49 +1,56 @@
...
+from tribler_core.components.base import Session
 from tribler_core.config.tribler_config import TriblerConfig
-from tribler_core.session import Session
+from tribler_core.start_core import components_gen
 
 BASE_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__), '..', '..', '..'))
 
 
-class IdleTribleRunner:
+async def idle_tribler_run():
```
devos50: **Question:** Besides the application tester, I don't think we have any Jenkins experiment that runs Tribler idle right? The would be a great validation test to see if we broken Gumby-Tribler support.

gumby/experiment.py
```diff
@@ -280,7 +280,8 @@ def get_changed_values(prev_dict, cur_dict):
 
         new_values, changed_values = get_changed_values(prev_dict, cur_dict)
         if changed_values:
-            self._stats_file.write('%.1f %s %s %s\n' % (time(), self.my_id, name, json.dumps(changed_values)))
```
devos50, line 283: **Remark:** This stuff wasn't even working apparently, good find 😁. We could consider adding a unit test for it?

scripts/build_virtualenv.sh
```diff
@@ -44,6 +44,7 @@ fi
 
 # If we compile for Python 3, we want to install a newer version since the version on the DAS6 is outdated.
 if [ ! -e ~/python3/bin/python3 ]; then
+    echo "Installing Python 3.9
```
devos50: **Issue:** Is this missing an ending `"`?

gumby/modules/tribler_module.py
```diff
@@ -126,19 +257,21 @@ def setup_initial_bootstrap_seeder(self):
                 fp.write(bytearray(random.getrandbits(8) for _ in range(bootstrap_size * 1024 * 1024)))
 
     @experiment_callback
-    def start_bootstrap_download(self):
-        self.session.start_bootstrap_download()
+    def start_bootstrap_download(self):  # obsolete?
```
devos50: **Suggestion:** I think we still have some settings related to the bootstrap download in Tribler but the feature seems to be disabled. Feel free to remove this logic from Gumby if it is obsolete in Tribler 👍 .

gumby/modules/tribler_module.py
```diff
@@ -273,7 +412,7 @@ def create_test_torrent(self, file_name, download_id, length):
         if not os.path.exists(file_name):
             self._logger.info("Creating torrent data file %s", file_name)
             with open(file_name, 'wb') as fp:
-                rand = Random()
+                rand = random.Random()
```
devos50, line 410: **Suggestion:** It might be a good idea to seed the randomness in our experiments to make them (more) deterministic. I'll create a separate issue for this 👍

devos50 (commented):
Note: my review policy in Gumby is less intensive compared to when reviewing code in Tribler. For Gumby, I primarily look at the experiment runs and expected outcomes (e.g., graphs, generated files etc.).

experiments/tribler_idle_run/tribler_idle_run.py
kozlovsky: Thanks for pointing me this; it was an old code. Initially, I fixed the code of `tribler_idle_run` to make it run without errors, but I did not change the overall structure. Now I rewrote `tribler_idle_run` to streamline the code.

experiments/tribler_idle_run/tribler_idle_run.py
kozlovsky: We can use `tribler_idle_run` in a validation test, but I'm not sure we should. We also have `tribler_module` that can be used in a scenario file. In my understanding, we usually use Gumby to run scenarios, and in a scenario that uses `tribler_module`, we can use all that `tribler_idle_run` allows, and more. Maybe later we can remove `tribler_idle_run` entirely in favor of `tribler_module`. But you have much more experience with Gumby, and perhaps there are use-cases where `tribler_idle_run` looks preferable to `tribler_module`.

gumby/experiment.py
kozlovsky: It's a good idea, but I think it is better to do it in a separate PR, as currently, we don't have a suitable unit test file for such a test. We can potentially test a lot of staff in `gumby/experiment.py` and other files, so adding proper unit tests looks like a solid chunk of work.

scripts/build_virtualenv.sh
kozlovsky: Oops... Fixed!

gumby/modules/tribler_module.py
kozlovsky: I removed it!

gumby/modules/tribler_module.py
kozlovsky: Sure

gumby/experiment.py
devos50: 👍

experiments/tribler_idle_run/tribler_idle_run.py
devos50: It might indeed be better to remove the `tribler_run_idle` script since there currently is no use case for it. The only use-case I can think of is if we want to run a version of Tribler on our compute cluster (or other infrastructure) that is as close as possible to the end-user configuration. But we currently do not need that so I would recommend to remove it. Should we do so in this PR or a separate one?

kozlovsky: I think it is better to remove `tribler_idle_run` in a separate PR. This way, it will be easier to revert the removing commit and return `tribler_idler_run` if it later becomes useful again.

devos50:
@kozlovsky please let me know when the experiments are fixed, so I can review/approve this PR 👍

kozlovsky:
@devos50 sure!

kozlovsky:
@devos50 the experiments are working:
https://jenkins-ci.tribler.org/job/validation_experiments/job/validation_experiments/1258/

devos50 (approved):
Looks good!

After merge, don't forget to change the cloning jobs of the validation experiments to point to the main branch again 👍

devos50:
Actually, it seems that the DAS experiment is not correctly generating build output anymore. This might be a DAS6 issue.

devos50:
retest this please

devos50:
Seems to be fixed now! (The version of `R` seems to have been updated)

PR #522, opened 2022-05-01
Renaming: das4_reserve_duration -> das_reserve_duration

kozlovsky:
Part of #524. Renames das4_reserve_duration to das_reserve_duration.

I'm not sure the DAS4_RESERVE_DURATION environment variable it is not specified somewhere in Jenkins jobs' configurations, so acceptiong and merging this PR may require some caution.


devos50 (approved):
As far as I'm aware, the `DAS4_RESERVE_DURATION` is not used explicitly in Jenkins jobs. We do set the `NODE_TIMEOUT` occasionally though.

PR #523, opened 2022-05-01
Renaming: das4_node_command -> das_node_command

kozlovsky:
Part of #524. Renames das4_node_command to das_node_command.

I'm not sure the DAS4_NODE_COMMAND environment variable it is not specified somewhere in Jenkins jobs' configurations, so acceptiong and merging this PR may require some caution.


devos50 (approved):
By default, the `DAS4_NODE_COMMAND` should run the `launch_scenario.py` script. This value is only different if we are not using scenario functionality. I'm not aware of any experiment where we are _not_ using scenario files to describe experiments. That said, this might break a few jobs if it modifies one of the variables you changed in your PRs, but the fix should be easy (change the Jenkins configuration).

